PR #42, opened 2020-03-04
Globalvars

phact:


virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_double/to_double/Expr.java
```diff
@@ -22,7 +26,14 @@ public Expr(String expr) {
 
     @Override
     public double applyAsDouble(double operand) {
+        ConcurrentHashMap<String, Object> gl_map = SharedState.gl_ObjectMap;
         HashMap<String, Object> map = SharedState.tl_ObjectMap.get();
+
+        // merge gl into tl, for duplicates use the value from tl
```
jshook: I'd like to see something more explicit here, since global and local state intermingling can be a source of difficult-to-diagnose errors.

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/globalvars/GlobalVarsWrapper.java
```diff
@@ -0,0 +1,20 @@
+package io.nosqlbench.engine.extensions.globalvars;
+
+import io.nosqlbench.virtdata.library.basics.core.threadstate.SharedState;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+public class GlobalVarsWrapper {
+    ConcurrentHashMap<String, Object> map = SharedState.gl_ObjectMap;
+
+    String test = "puppies";
```
jshook: too many puppies

phact: Whoops didn't mean to commit this class, removing.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_double/to_double/Expr.java
phact: I'm open to ideas.
I guess we could have a TL Expr and a Global Expr?

I imagine there are probably some cases where you may want to pull from both but we could force folks to compose for that.

jshook: I think we should at least make the naming clear in terms of `local` `global` and perhaps a `combined` or `all` version. Changing from one set (or superset) to another of included variables should be a simple change of a function name. I favor this over having users have to compose data sources as an inner function. We can make the extant "anonymous" versions throw a logger warning that says "Use one of the other clearly named functions instead".

What do you think?

PR #45, opened 2020-03-04
Async docs

phact:
This improves performance for slow networks by pulling all the md's at once asyncronously.

The docs are still broken because content is missing so I haven't been able to do a full test.


phact:
Merged in master w/ docs content.
Fixed some bugs, should be good to go.

phact:
Also added the static site for gh pages into this same PR

docsys/src/main/node/docsys/update.sh
```diff
@@ -2,4 +2,11 @@
 rm -rf dist .nuxt
 npm run generate
 rm -rf ../../resources/docsys-guidebook
-mv dist ../../resources/docsys-guidebook
+cp -r dist ../../resources/docsys-guidebook
```
jshook: This works for now, but we have multiple doc sources that are all loaded automatically in the run time.
A future version of our static site generation should use `nuxt generate` with an optional environment variable that tells the dynamic route mechanism how to find and cache all the pages exposed by the running docserver.

We can get by for now by updating this script, but the number of `../` is concerning.

phact: works for me

PR #66, opened 2020-03-16
[issue-61] Use OSHI library for getting system information

### LINKED ISSUE #61: CpuInfo class is too Linux dependent
Right now, `CpuInfo` and `CpuInfo.ProcDetails` classes are too Linux dependent as they are reading `/proc/cpuinfo`, and other files from `/proc/` that aren't available on the Mac OS, and other platforms...  We need to decrease dependency on the low level stuff like this, by maybe using https://github.com/oshi/oshi, or something like...

alexott:
This fixes #61


phact:
lgtm +1

PR #75, opened 2020-03-17
Issue 67

phact:


engine-cli/src/main/java/io/nosqlbench/engine/cli/EBCLIOptions.java
```diff
@@ -313,10 +321,66 @@ private void parse(String[] args) {
...
+            for (String parameter: cmdArray) {
+                if (parameter.contains("=")){
+                    if ( !parameter.contains("TEMPLATE(") && !parameter.contains("<<")) {
+                        String[] paramArray = parameter.split("=");
+                        paramMap.put(paramArray[0], paramArray[1]);
+                    }
+                }
             }
+
+            StrSubstitutor sub1 = new StrSubstitutor(paramMap, "<<", ">>", '\\', ",");
```
jshook: This is handled prior to yaml loading, you can remove it.

engine-cli/src/main/java/io/nosqlbench/engine/cli/EBCLIOptions.java
```diff
@@ -313,10 +321,66 @@ private void parse(String[] args) {
...
+            StrSubstitutor sub2 = new StrSubstitutor(paramMap, "TEMPLATE(", ")", '\\', ",");
+
+            cmd = sub2.replace(sub1.replace(cmd));
+
+            // Is there a better way to do this than regex?
```
jshook: I do think this will break if there is a case for spaces within parameter names or values, but otherwise, it should be ok.

phact:
Docs are still in the works for this btw

jshook:
I manually merged this up and pushed updates, so we can close this one.

PR #85, opened 2020-03-19
use Java 12 everywhere

alexott:
the `docsys` module did use Java 12 for it, while the rest was compiled for Java 11.  As
we're packing with Java 12 anyway, let switch to it for everything.

### DIFF
--- a/mvn-defaults/pom.xml
+++ b/mvn-defaults/pom.xml
@@ -11,7 +11,7 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <javadoc.name>nosqlbench</javadoc.name>
 
-        <java.target.version>11</java.target.version>
+        <java.target.version>12</java.target.version>
         
         <!-- properties for package versions -->
         <antlr4.version>4.7.2</antlr4.version>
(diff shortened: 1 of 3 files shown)


mvn-defaults/pom.xml
```diff
@@ -11,7 +11,7 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <javadoc.name>nosqlbench</javadoc.name>
 
-        <java.target.version>11</java.target.version>
+        <java.target.version>12</java.target.version>
```
jshook: We may need the source and release properties as well.
The release property is a newly supported way to specify "source and target", but older javac versions won't recognize it, thus why all three properties were initially provided.

jshook: The reason to support recognized flags for older javac versions is so that developers get useful error messages when they have an older JDK installed, which is not uncommon.

alexott: All 3 are already set to the value of this variable: https://github.com/nosqlbench/nosqlbench/blob/master/mvn-defaults/pom.xml#L453 - I can maybe rename it to the `java.version` ? Or you want that we have a flexibility in defining different source/target/release?

jshook: As long as they are all set, it works. I missed that detail in the review. I think they should be set to the same value in general.

PR #88, opened 2020-03-20
[issue-77] build-nb-appimage should happen only on linux

### LINKED ISSUE #77: build-nb-appimage should happen only on linux
Right now, `build-nb-appimage` is trying to execute on Mac OS as well, and fail because it uses incorrect image.

alexott:
this fixes #77


alexott:
I've also tested it with manual build on empty Linux machine...

phact:
LGTM +1

PR #110, opened 2020-04-06
reconnectpolicy

phact:


activitytype-cql/src/main/java/io/nosqlbench/activitytype/cql/core/CQLOptions.java
```diff
@@ -83,6 +84,24 @@ public static RetryPolicy retryPolicyFor(String spec) {
         return retryPolicy;
     }
 
+    public static ReconnectionPolicy reconnectPolicyFor(String spec) {
+       if(spec.startsWith("exponential(")){
+           String argsString = spec.substring(12);
```
jshook, line 89: For suffix extractions, I like to use the pattern x.substring(y.length()), since it is less error prone.
We won't miss the cycles spent on configuration, but we might miss them more in bug fixes when names change.
No need to change the PR

PR #118, opened 2020-04-15
Ft Raw query

Pierrotws:
This feature allows user to use data bindings with simple statement. Statement is  generated by 
RawValueBinder :
context is the parsed statement for driver string splitted by "?"
binder merge context using values[]
This Binder is selected if both prepared=false and raw=true
Use of an ENUM (prepared/raw/parametrized) may be a little bit cleaner but breaks ascendant compatibility


activitytype-cql/src/main/java/io/nosqlbench/activitytype/cql/core/CqlActivity.java
```diff
@@ -204,6 +205,17 @@ private void initSequencer() {
...
+                cl.ifPresent((conlvl) -> {
+                    psummary.append(" consistency_level=>").append(conlvl);
+                });
+                serial_cl.ifPresent((scl) -> {
+                    psummary.append(" serial_consistency_level=>").append(scl);
+                });
+                idempotent.ifPresent((i) -> {
+                    psummary.append(" idempotent=>").append(i);
+                });
+                template = new ReadyCQLStatementTemplate(getSession(), stmtForDriver, ratio, parsed.getName(), cl, serial_cl, idempotent);
```
jshook, line 218: Can you use the form with setters as above so the code is easier to follow?

activitytype-cql/src/main/java/io/nosqlbench/activitytype/cql/statements/binders/RawValueBinder.java
```diff
@@ -0,0 +1,48 @@
...
+import io.nosqlbench.virtdata.core.bindings.ValuesArrayBinder;
+
+import java.util.Optional;
+
+/**
+ * This binder is only meant to be used to directly generate final SimpleStatement query
+ * without use of preparedStatement. This should be avoided any time possible since it's far more
+ * less optimized than use of PreparedStatement. Only Use it when PreparedStatement is not possible
+ */
+public class RawValueBinder
```
jshook, line 15: We should probably use a StringBindingsTemplate to do this for us as a variant of the ReadyCQLStatement.

Pierrotws:
My binder was called RawValue Binder and was triggered by parameter raw=true which is a little bit confusing.

Previous version of SimpleStatementValuesBinder document specified that this behaviour should be implemented here.
So my code was backported to directly be in SimpleStatementValuesBinder.
Context is now SimpleStatement instead of String[].
Behaviour is activated when prepared=false and parametrized=true

activitytype-cql/src/main/java/io/nosqlbench/activitytype/cql/statements/binders/SimpleStatementValuesBinder.java
```diff
@@ -1,19 +1,55 @@
...
     public Statement bindValues(SimpleStatement context, Object[] values) {
-        return new SimpleStatement(context.getQueryString(), values);
+        String query = context.getQueryString();
+        if(parametrized) {
+            String[] splits = query.split("\\?");
+            assert splits.length == values.length+1;
+            StringBuilder sb = new StringBuilder();
+            sb.append(splits[0]);
+            for(int i = 1; i < splits.length; i++) {
+                sb.append(values[i - 1]);
```
jshook, line 32: For string fields, this may require quote injection for non-primitive numeric types, etc. Since the non-parameterized form assumed the same syntax as used in prepared statements, quotes are not safe to assume. We should make them work co-compatibly.

PR #138, opened 2020-05-08
CSVRecord null checks

phact:


jshook:
@phact Can you update this branch to be mergable again?

phact:
Yep

PR #153, opened 2020-05-27
Add MongoDB support

justinchuch:
Starting with the basics:

- `insert`
- `find`

### DIFF
--- a/driver-mongodb/src/main/java/io/nosqlbench/driver/mongodb/MongoAction.java
+++ b/driver-mongodb/src/main/java/io/nosqlbench/driver/mongodb/MongoAction.java
@@ -1,0 +1,64 @@
+package io.nosqlbench.driver.mongodb;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.codahale.metrics.Timer;
+import com.mongodb.ReadPreference;
+import com.mongodb.client.MongoDatabase;
+import io.nosqlbench.engine.api.activityapi.core.SyncAction;
+import io.nosqlbench.engine.api.activityapi.planning.OpSequence;
+import org.bson.Document;
+import org.bson.conversions.Bson;
+
+public class MongoAction implements SyncAction {
+
+    private final static Logger logger = LoggerFactory.getLogger(MongoAction.class);
+
+    private final MongoActivity activity;
+    private final int slot;
+
+    private OpSequence<ReadyMongoStatement> sequencer;
+
+    public MongoAction(MongoActivity activity, int slot) {
+        this.activity = activity;
+        this.slot = slot;
+    }
+
+    @Override
+    public void init() {
+        this.sequencer = activity.getOpSequencer();
+    }
+
+    @Override
+    public int runCycle(long cycleValue) {
+        ReadyMongoStatement rms;
+        Bson queryBson;
+        try (Timer.Context bindTime = activity.bindTimer.time()) {
+            rms = sequencer.get(cycleValue);
+            queryBson = rms.bind(cycleValue);
+
+            if (activity.isShowQuery()) {
+                logger.info("Query(cycle={}):\n{}", cycleValue, queryBson);
+            }
+        }
+
+        for (int i = 0; i < activity.getMaxTries(); i++) {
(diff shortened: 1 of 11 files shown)


driver-mongodb/src/main/java/io/nosqlbench/driver/mongodb/MongoAction.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public int runCycle(long cycleValue) {
+        ReadyMongoStatement rms;
+        Bson queryBson;
+        try (Timer.Context bindTime = activity.bindTimer.time()) {
+            rms = sequencer.get(cycleValue);
+            queryBson = rms.bind(cycleValue);
+
+            if (activity.isShowQuery()) {
+                logger.info("Query(cycle={}):\n{}", cycleValue, queryBson);
```
jshook, line 42: We might want to add a warning on this when it is enabled and cycles>1000 total, similar to how we do with the CQL activity.

driver-mongodb/src/main/java/io/nosqlbench/driver/mongodb/MongoAction.java
```diff
@@ -0,0 +1,64 @@
...
+            try (Timer.Context executeTime = activity.executeTimer.time()) {
+                MongoDatabase database = activity.getDatabase();
+                ReadPreference readPreference = rms.getReadPreference();
+                Document resultDoc = database.runCommand(queryBson, readPreference);
+
+                double ok = resultDoc.getDouble("ok");
+                activity.resultSetSizeHisto.update(resultDoc.getInteger("n", 0));
+                return ok == 1.0d ? 0 : 1;
+            } catch (Exception e) {
+                logger.error("Failed to runCommand {} on cycle {}, tries {}", queryBson, cycleValue, i, e);
```
jshook, line 57: As a future enhancement, we will want to capture total results and successful results as separate metrics, requiring a slight refactoring of the exception handling here.

driver-mongodb/src/main/resources/activities/mongodb-basic.yaml
```diff
@@ -0,0 +1,57 @@
+# nb -v run driver=mongodb yaml=mongodb-basic connection=mongodb://127.0.0.1 database=testdb tags=phase:rampup
```
jshook, line 1: There is now a `description` property supported in YAMLs that hasn't been documented yet. You might want to add a `description:` to this yaml.

driver-mongodb/src/main/resources/mongodb.md
```diff
@@ -0,0 +1,3 @@
+# MongoDB Driver
+
+This is a driver for MongoDB.
```
jshook, line 3: We should document all the parameters that are special to this activity type.

pom.xml
```diff
@@ -68,6 +69,16 @@
                 <module>driver-cqld4</module>
             </modules>
         </profile>
+        <profile>
+            <id>with-mongodb</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
```
jshook, line 75: This needs to be set to true to work with the module inclusion above, in which case the lower module inclusion also applies. I suggest we just make it on by default now that it is building and working.

driver-mongodb/src/main/resources/activities/mongodb-basic.yaml
justinchuch: Updated. :ok_hand:

pom.xml
justinchuch: Updated. :ok_hand:

driver-mongodb/src/main/java/io/nosqlbench/driver/mongodb/MongoAction.java
justinchuch: Added `resultSuccess` metric, renamed the metric `execute` to `result`.

driver-mongodb/src/main/resources/mongodb.md
justinchuch: Updated. :ok_hand:

PR #154, opened 2020-05-29
[WIP] Pull the substride timing mechanism from engine block

pingtimeout:
The should contain the changes [that were applied on engineblock](https://github.com/engineblock/engineblock/compare/f1199b79a5ac9e62a2f3722156588dae181a747f...substride?w=1).  There is an extra try/finally to have a consistent try-while combo for both strides and substrides


jshook:
This was done because the master branch was deleted in lieu of the main branch. I can still get to the changes to pull them in.

PR #194, opened 2020-08-13
counter fix

phact:


jshook:
@phact Is this fix still valid?

jshook:
I manually applied the fix described in the parser logic.

If the changes to the dependencies were important, please let me know. Otherwise I'll close this PR.

PR #201, opened 2020-09-08
Improve SSLKsFactory

justinchuch:
Use `kspass` (i.e. keystore password) as the `keyPassword` if `keyPassword` is not specified.


jshook (approved):
LGTM, this will go out with the next release tonight.

PR #207, opened 2020-09-23
fix WeightedLongs function: member values not allocated

ncarvind:
Allocate memory for member variable values


jshook (approved):
Thanks for the fix!

PR #208, opened 2020-09-23
Changes for dealing with avro serialization of kafka messages

ncarvind:
Change KafkaActivity to derive from SimpleActivity for better flow.

I committed a change for another issue in this branch by mistake and had to revert it.

### DIFF
--- a/driver-kafka/pom.xml
+++ b/driver-kafka/pom.xml
@@ -27,6 +27,20 @@
             <version>2.0.0</version>
         </dependency>
 
+        <!-- https://mvnrepository.com/artifact/org.apache.avro/avro -->
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+            <version>1.10.0</version>
+        </dependency>
+
+        <!-- https://mvnrepository.com/artifact/io.confluent/kafka-avro-serializer -->
+        <dependency>
+            <groupId>io.confluent</groupId>
+            <artifactId>kafka-avro-serializer</artifactId>
+            <version>5.5.1</version>
+        </dependency>
+
         <dependency>
             <groupId>io.nosqlbench</groupId>
             <artifactId>engine-api</artifactId>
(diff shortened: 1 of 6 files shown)


driver-kafka/pom.xml
```diff
@@ -62,7 +76,17 @@
         </dependency>
 
     </dependencies>
-
+    <repositories>
```
jshook, line 79: It's best to avoid adding vendor-specific repositories to the project where we can. Is io.confluent:kafka-avro-serializer  available in any mainstream repository?

jshook: I do recognize that we'll need to make exceptions to this where necessary, but the goal is to avoid adding dependencies to the build where possible. As a local dependency in the kafka module, it's not a huge impact.

ncarvind: @jshook Redhat maven repository seems to have an old version which is not working for me. Couldn't find it anywhere else.

driver-kafka/src/main/java/com/datastax/ebdrivers/kafkaproducer/KafkaStatement.java
```diff
@@ -0,0 +1,141 @@
...
+                throw new RuntimeException(e);
+            }
+        }
+
+        return statement;
+    }
+
+    private Object bindValue(long cycle) {
+        Object statement = bindings.bind(cycle);
+        if (valueSerializerClass != null &&
```
jshook, line 118: Same here, it would be nice if there were always a defined serializer to call via an interface.

driver-kafka/src/main/java/com/datastax/ebdrivers/kafkaproducer/KafkaStatement.java
```diff
@@ -0,0 +1,141 @@
...
+        try {
+            producer = new KafkaProducer<>(props);
+        } catch (Exception e) {
+            logger.error("Error constructing kafka producer", e);
+        }
+    }
+
+    private Object bindKey(long cycle) {
+        Object statement = key;
+        if (keySerializerClass != null &&
```
jshook, line 104: It would be nice to indirect this through an always-defined function at some point.

jshook (approved):
This all makes sense to me. Thanks for the contribution.
Can I ask you to add some docs to the kafka.md that explain the set of options to users, including what is there before?

jshook (approved):

driver-kafka/pom.xml
jshook: Understood. We'll leave it in here, as this is incremental functionality and it is module and protocol-specific. I don't see a better alternative. We need to add something about avoiding repo dependencies into the dev guide just to clarify the reasoning, though.

ncarvind:
Yup couple things still to do but didnt have time
1. Figure out how to let users specify key thats parameterized.
2. Complete documentation.
3. Address ur concerns above.

PR #251, opened 2021-01-25
CockroachDB driver

szimmer1:
This branch is based on @jhatcher9999's branch https://github.com/jhatcher9999/nosqlbench/commit/6d2930f978f2fd9e71ffb7b4fa17f55463996f32

The design attempts to conform to the basic requirements in [driver_standards.md](https://github.com/nosqlbench/nosqlbench/blob/main/devdocs/devguide/driver_standards.md). It differs significantly from the other driver designs: much of the logic is implemented by an abstract JDBCActivity, and other SQL drivers should inherit from and define their specific behavior (e.g. DataSource initialization, retry-able error codes, etc).

TODO

- Unit tests at least for JDBCAction

### DIFF
--- a/driver-cockroachdb/pom.xml
+++ b/driver-cockroachdb/pom.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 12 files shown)


jshook (approved):
This looks like a good direction for the JDBC family of drivers. Thanks for jumpin in @szimmer1 !
I'll merge this up into main and plan to announce it in the list of NB4 major enhancements as long as it gets some tree shaking before then. If not, we can describe it as experimental pending testing.

PR #254, opened 2021-01-28
Add support for specifying a haproxy header for new cql connections

tjake:


jshook (approved):
Thanks for adding this!

PR #255, opened 2021-01-28
Add ByteBufferSizedHashed

Pierrotws:
Like HashedToByteBuffer, but param is a size function instead of a fixed length


jshook (approved):
Looks good!

PR #272, opened 2021-02-26
remove parameters

pushkala-datastax:


driver-http/src/main/resources/activities/baselines/http-iot.yaml
```diff
@@ -27,8 +27,6 @@ bindings:
 blocks:
   - tags:
       phase: schema
-    params:
-      prepared: false
```
jshook, line 31: This params property sets the default for any statements in this block, which is only DDL. If the statements are prepared (true by defult), then they will be rejected, as prepared statements need a schema to be checked against.

jshook: Maybe we can find a better way to handle this in terms of defaults and user feedback.

jshook: Can we close this PR and discuss the issue in more detail?

pushkala-datastax:
Yes please.

driver-http/src/main/resources/activities/baselines/http-iot.yaml
pushkala-datastax: yes plz

PR #281, opened 2021-03-12
Changed the schema creation to cql to be consistent with cql baselines.

jeromatron:
We're not trying to test schema creation with the alternative Stargate APIs at this point, so may as well make schema creation consistent across different baselines for now.


jshook (approved):
We need to make this work only with http at some point. I realize there is a delta on the API side that makes this difficult right now, but we should make "http" mode work self-standing ASAP. Thank for the updates along the way!

PR #283, opened 2021-03-17
Enhancements on Pulsar Driver

eolivelli:
- handle correctly Consumer configuration (and remove "timeout" to the configuration passed down to Pulsar, because it crashes the client)
- use NBErrorHandler for PulsarActivity


jshook (approved):
Excellent work and quick turn-around since our discussion.

PR #285, opened 2021-03-17
Adding weighted hosts and flexible url

jeromatron:
Making the hosts into a weighted string to allow for round robin and weighted host rotation with Stargate hosts.  Also parameterizing the protocol (http|https) and the path prefix so that this baseline can be run against Stargate directly or against Astra.


jshook:
Thanks

PR #286, opened 2021-03-22
Pulsar: add bytes counters

eolivelli:
- Add "bytes" counter in order to track the number for bytes written to Pulsar and see the thoughtput in bytes and not in ops/sec.
- Fix dependencies

### DIFF
--- a/driver-pulsar/pom.xml
+++ b/driver-pulsar/pom.xml
@@ -27,7 +27,7 @@
 
         <dependency>
             <groupId>org.apache.pulsar</groupId>
-            <artifactId>pulsar-client</artifactId>
+            <artifactId>pulsar-client-original</artifactId>
             <version>${pulsar.version}</version>
         </dependency>
 
@@ -75,7 +75,7 @@
         <dependency>
             <groupId>org.apache.avro</groupId>
             <artifactId>avro</artifactId>
-            <version>1.10.1</version>
+            <version>1.9.1</version>
         </dependency>
 
     </dependencies>
--- a/driver-pulsar/pom.xml
+++ b/driver-pulsar/pom.xml
@@ -27,7 +27,7 @@
 
         <dependency>
             <groupId>org.apache.pulsar</groupId>
-            <artifactId>pulsar-client-original</artifactId>
+            <artifactId>pulsar-client</artifactId>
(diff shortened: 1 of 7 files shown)


eolivelli:
@jshook is this the right way to add a counter to see the "thoughtput" ? or should I use some other kind of metric ?

driver-pulsar/pom.xml
```diff
@@ -27,7 +27,7 @@
 
         <dependency>
             <groupId>org.apache.pulsar</groupId>
-            <artifactId>pulsar-client</artifactId>
+            <artifactId>pulsar-client-original</artifactId>
```
jshook, line 30: What does this change do?

eolivelli: "pulsar-client" is a JAR that contains all of the dependencies of Pulsar, shaded, it is like an uber jar.

eolivelli: without this change IntelliJ messes up.

I didn't check how we package drivers. is it a problem if this jar adds its own third party deps to the NB build ?

eolivelli: @jshook I can revert these changes to pom.xml

eolivelli:
by the way I cannot see this new metric after the execution of NB.
is there is something more to do in order to see the metric published into the CSV files ?

driver-pulsar/pom.xml
jshook: The artifacts need to be resolvable in a public repository or it will break the build for everybody.

jshook:
> @jshook is this the right way to add a counter to see the "thoughtput" ? or should I use some other kind of metric ?

I think it is suitable for tracking the cumulative payload size sent, but it will not give you any direct insights of payloads over time without doing your own derivatives, nor will it provide histograms.
You could also use a histogram if you want more texture of the data over time. It might make sense to keep the counter and a histogram both, since the counter is lightweight and provides a baseline signal as is.

driver-pulsar/pom.xml
jshook: I see that this is also a public artifact. No issues on this from me.

jshook (approved):
This look ok. It might be useful to hold these metric objects in the ready op at some point.

PR #289, opened 2021-03-29
adding graphql schema first scenarios

ivansenic:
Here are some guides for the reviewer, there is a list of `TODO`s in each file, but I'll try to mention the most important ones:

* I tried to replicate as must as possible the REST scenarios, however schema creation is always gql-first
* default keyspace is different in each scenario, as we have schema per keyspace in GQL
* tabular scenario
   * the rest equivalent of this uses only `part_read` when reading, but this is not possible with GraphQL API at the state we have it in the moment, as both primary keys need to be part of the querying
   * also I am not sure how `limit` would play a role due to this requirement
* timeseries
   * I can not replicate the compression and compaction as I am using graphql api to create the table (and this is not supported)
   * when I get `-Infinity` for the `sensor_name` the API request fails to parse the content, but respond to `200`.. however it would be good to skip those requests
   * paging is still not implemented in the requests, I guess it fails to the default page size of 10
   
In order to test this, you must be on the Stargate PR https://github.com/stargate/stargate/pull/634.. APIs called here are not released yet.. Example run:

```
./nb -v run driver=http yaml=./driver-http/src/main/resources/activities/graphql-schema-first/http-graphql-timeseries.yaml tags=phase:schema host=127.0.0.4 stargate_host=127.0.0.4 auth_token=$AUTH_TOKEN 
./nb -v run driver=http yaml=./driver-http/src/main/resources/activities/graphql-schema-first/http-graphql-timeseries.yaml tags=phase:rampup host=127.0.0.4 stargate_host=127.0.0.4 auth_token=$AUTH_TOKEN cycles=1000 threads=auto
./nb -v run driver=http yaml=./driver-http/src/main/resources/activities/graphql-schema-first/http-graphql-timeseries.yaml tags=phase:main host=127.0.0.4 stargate_host=127.0.0.4 auth_token=$AUTH_TOKEN cycles=10000 threads=auto
```

### DIFF
--- a/driver-http/src/main/resources/activities/graphql-schema-first/http-graphql-keyvalue.yaml
+++ b/driver-http/src/main/resources/activities/graphql-schema-first/http-graphql-keyvalue.yaml
@@ -1,0 +1,114 @@
+# nb -v run driver=http yaml=http-graphql-keyvalue tags=phase:schema host=my_stargate_host stargate_host=my_stargate_host auth_token=$AUTH_TOKEN
+
+# TODO 
+# - /graphqlv2 will change in future, adapt when needed
+# - do we need a truncate schema / namespace at the end
+
+description: |
+  This workload emulates a key-value data model and access patterns.
+  This should be identical to the cql variant except for:
+  - Schema creation GraphQL first, we don't use cql and thus can only create schema with limited options.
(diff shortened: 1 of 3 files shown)


jshook (approved):
Thanks!

PR #291, opened 2021-03-30
Allow to use  the Pulsar Producer Async API in order to generate the maximum load

eolivelli:
This patch allows the PulsarProducerOp to use the Pulsar Async send API.

The key point is that we are not blocking to the result of the send, the thread will block automatically in case that the internal queue is full (this is the blockIfQueueFull configuration parameter of the Pulsar Producer).

The producer internally batches writes and performs all of the available optimisations implemented on the PulsarClient.
It is important to let Pulsar client do this work and to not perform batching at application level, because we want to leverage all of the features of Pulsar without reimplementing them.

With this patch we can make NB write as fast as possible to Pulsar (x10 throughput on my test env) , as the pulsar-perf tool (builtin in the Pulsar distribution) does.

Metrics and error reporting:
we can still report the execution time metrics, but in order to do it I had to defer the invokation of Timer.Context#close to the internal Pulsar callback mechanism.
As the error will be reported on a CompletableFuture I had to introduce a little trick to fail NB execution in case of any error.

This patch does not match the NB Async mode, but it is enough in order to create enough load on Pulsar and also to be able to leverage the internal batching features of the Pulsar client.


eolivelli:
@jshook can we also talk about this patch, together with my problem of dumping internal Pulsar client metrics ?

eolivelli:
@yabinmeng you may be interested as well

yabinmeng:
@eolivelli yes I'm very interested in :)

Just a heads up that this change will be conflict with my latest PR of adding a new Admin API

eolivelli:
@yabinmeng your PR has already been merged, this patch is on top of that

driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/PulsarOp.java
```diff
@@ -3,6 +3,12 @@
 /**
  * Base type of all Pulsar Operations including Producers and Consumers.
  */
-public interface PulsarOp extends Runnable {
+public interface PulsarOp {
```
jshook, line 6: We want to keep this implementing Runnable directly, I think. I'll merge it for now to support your testing, but we should discuss the API strategy for this.

jshook (approved):
Thanks

eolivelli:
@jshook I have resolved the conflicts and tested again the patch.
it works well :-)

PR #295, opened 2021-04-01
Use CSVFormat.Default to handle rfc4180 CSVs

phact:


.github/workflows/release.yml
```diff
@@ -162,6 +162,7 @@ jobs:
         run: scripts/release-perform.sh
         continue-on-error: true
         env:
+          RELEASE_BRANCH_PATTERN: "release"
```
jshook, line 165: I'll set this to main after merging

jshook (approved):

.github/workflows/release.yml
phact: Thanks!

PR #303, opened 2021-04-21
Pulsar: first prototype of transaction support

eolivelli:
Introduce "use_transaction" parameter for Produce Op and ConsumeOp.
Add metrics about starting the transaction and commit the transaction.

Support for the Producer is 100% working both in sync and in async mode.
Support for Consumer is only a prototype as we need to separate the time spent for "receive" and for "commit".


yabinmeng:
Looks good to me. @eolivelli

jshook:
will do

jshook (approved):
This looks good to me. We should probably start breaking out different types of PulsarOps, as it seems that the current one is becoming layered up with conditional behaviors. The "mappers" can produce any type that works, and it would be better to have smaller more specific op implementations.

PR #305, opened 2021-04-21
Fix graphql timeseries read query

dougwettlaufer:
Fixing bad copy paste in the graphql query of the read request in the timeseries workload.


jshook (approved):
thanks!

PR #306, opened 2021-04-21
Use http requests when creating schema for rest and graphql

dougwettlaufer:
Not updating the timeseries workloads since they require special table options not currently exposed by stargate.

cc: @jeromatron @ivansenic


jshook (approved):
Nice!

jshook:
@all-contributors please add @jeromatron for tests

PR #312, opened 2021-04-30
Pulsar JMS driver type name change and package name typo fix

yabinmeng:


eolivelli:
@jshook @yabinmeng I missed this work was happening.

I have two points:
- Pulsar JMS client is not on Maven Central yes, so users out of Datastax are not able to build NB, correct?
- JMS is a general purpose API, I would call this module only JMS and let the name of the Connection Factory class name configurable
This way:
- you do not depend at build time on Pulsar JMS
-  we will be able to test other JMS providers and make comparisons

jshook:
If the artifacts are not on Maven central, then we should roll back this driver to a feature branch until it is. All NB users should be able to compile NB.

I agree with naming this JMS and letting the JMS config and late bindings do the rest.

yabinmeng:
@eolivelli thanks for the comment. @jshook I will revert PR 311 and 312

PR #313, opened 2021-05-01
Revert "Pulsar JMS driver type name change and package name typo fix"

yabinmeng:
Reverts nosqlbench/nosqlbench#312


yabinmeng:
@jshook  Looks like I can't revert 311 at this point. After 312 is reverted, I will retry reverting 311

PR #320, opened 2021-05-21
Add JSON API keyvalue workload

EricBorczuk:
In order to run benchmarks against the Documents API, I created this modified workload for key-value that I think can be included alongside graphql workloads for the HTTP driver!


jshook (approved):
Looks good

PR #322, opened 2021-05-29
Adds postgres-basic workload and scenarios

szimmer1:
Because the cockroachdb driver is actually using the postgresql JDBC driver, it is compatible with postgres.

This is a sample workload and scenario for postgres.


jshook (commented):
Thanks!

PR #328, opened 2021-06-10
avoid decoding msgBody twice in JmsMsgSendOp

XN137:
small improvement i noticed while reading through https://github.com/nosqlbench/nosqlbench/pull/311

not sure if this has any positive impact on the driver's performance.


eolivelli (approved):
LGTM

good catch

PR #329, opened 2021-06-10
simplify cache key creation in pulsar + jms driver

XN137:
i was looking into whether the use of the default system encoding in `.getBytes()` in the `static String encode(String... strings)` helper methods was problematic or not.
turns out it isnt, because the result is only used as a caching key in a `ConcurrentHashMap`, however this also means that the base64 encoding is completely unnecessary imo.

note that previously only parts where `StringUtils.isBlank(part)` returns false were joined, while now the parts are joined always as `"null"` or `""`.
this change in behavior imo is actually preferrable since `encode("foo", null)` and `encode(null, "foo")` used to return the same result, whereas now they would return `foo::null` and `null::foo`, which seems more deterministic (no unwanted cache collisions?)

### DIFF
--- a/driver-jms/src/main/java/io/nosqlbench/driver/jms/JmsActivity.java
+++ b/driver-jms/src/main/java/io/nosqlbench/driver/jms/JmsActivity.java
@@ -116,13 +116,16 @@ public class JmsActivity extends SimpleActivity {
         );
     }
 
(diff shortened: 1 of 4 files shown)


eolivelli (approved):
LGTM

jshook (approved):

PR #332, opened 2021-06-14
move virtdata-api tests to correct folder

XN137:
i noticed that most tests of `virtdata-api` were not getting executed since they were in the wrong folder.

fixed via these cmds:
```
git mv virtdata-api/src/test/io/nosqlbench/virtdata/{annotations,testmappers,util} virtdata-api/src/test/java/io/nosqlbench/virtdata
git mv virtdata-api/src/test/io/nosqlbench/virtdata/core/* virtdata-api/src/test/java/io/nosqlbench/virtdata/core
```

see inline comment for a test that is actually failing (we ignore it for now since i saw no easy solution).

tested locally with `mvn -T 0.5C clean package`


virtdata-api/src/test/java/io/nosqlbench/virtdata/core/templates/ParsedTemplateTest.java
```diff
@@ -40,6 +39,7 @@ public void testShoudlMatchCurlyBraces() {
         assertThat(pt.getExtraBindings()).hasSameElementsAs(bindings.keySet());
     }
 
+    @Ignore("currently not working correctly")
```
XN137, line 42: i saw no easy solution for this failing test... seems like the `?` handling is broken

jshook: Thanks! I'll take this PR and fix up anything that is no longer valid.

PR #336, opened 2021-06-17
fix errors found by errorprone

XN137:
i locally ran the build with errorprone:
http://errorprone.info/index

this found a few bugs and quite a bunch of suspicous code.
i will comment the most important ones inline, please review carefully for the potential impact of the fix.

among them:
http://errorprone.info/bugpattern/ConstantOverflow
http://errorprone.info/bugpattern/EqualsHashCode
http://errorprone.info/bugpattern/JUnit4TestNotRun
http://errorprone.info/bugpattern/OptionalMapUnusedValue
http://errorprone.info/bugpattern/SelfAssignment
http://errorprone.info/bugpattern/UnusedCollectionModifiedInPlace

http://errorprone.info/bugpattern/ImmutableEnumChecker
http://errorprone.info/bugpattern/StreamResourceLeak

tested locally via:
```
mvn -T 0.5C clean package
```

in the future we can add errorprone to the build by default... i just need to figure out what other annotationprocessors there are apart from lombok?

### DIFF
--- a/docsys/src/main/java/io/nosqlbench/docsys/endpoints/DocsysMarkdownEndpoint.java
+++ b/docsys/src/main/java/io/nosqlbench/docsys/endpoints/DocsysMarkdownEndpoint.java
@@ -180,12 +180,6 @@ public class DocsysMarkdownEndpoint implements WebServiceObject {
     }
 
     private void enable(Set<String> enabled) {
-
-        Set<String> toEnable = new HashSet<>();
-        if (this.enables !=null) {
-            toEnable.addAll(this.enables);
-        }
-
         for (DocsNameSpace nsinfo : docsinfo) {
             // add namespaces which are neither enabled nor disabled to the default group
             if (nsinfo.isEnabledByDefault()) {
@@ -204,6 +198,4 @@ public class DocsysMarkdownEndpoint implements WebServiceObject {
             this.enabled = disabled.remove(enabled);
         }
     }
-
-
 }
--- a/driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/datamappers/functions/to_daterange/DateRangeFunc.java
+++ b/driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/datamappers/functions/to_daterange/DateRangeFunc.java
@@ -75,13 +75,12 @@ public class DateRangeFunc implements LongFunction<DateRange> {
         this.upper = upper;
     }
 
-
     @Override
     public DateRange apply(long value) {
         Date lowerDate = new Date(lower.applyAsLong(value));
         DateRange.DateRangeBound lower = DateRange.DateRangeBound.lowerBound(lowerDate,precision);
         Date upperDate = new Date(upper.applyAsLong(value));
-        DateRange.DateRangeBound upper = DateRange.DateRangeBound.upperBound(lowerDate,precision);
+        DateRange.DateRangeBound upper = DateRange.DateRangeBound.upperBound(upperDate,precision);
         DateRange dateRange = new DateRange(lower, upper);
         return dateRange;
     }
--- a/driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/statements/core/CQLSessionCache.java
+++ b/driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/statements/core/CQLSessionCache.java
@@ -320,7 +320,7 @@ public class CQLSessionCache implements Shutdownable {
         Cluster cl = builder.build();
 
         // Apply default idempotence, if set
-        activityDef.getParams().getOptionalBoolean("defaultidempotence").map(
+        activityDef.getParams().getOptionalBoolean("defaultidempotence").ifPresent(
             b -> cl.getConfiguration().getQueryOptions().setDefaultIdempotence(b)
         );
 
--- a/driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/statements/core/CQLStatementDefParser.java
+++ b/driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/statements/core/CQLStatementDefParser.java
@@ -12,11 +12,8 @@ public class CQLStatementDefParser {
     private final static Logger logger = LogManager.getLogger(CQLStatementDefParser.class);
     //    private final static Pattern templateToken = Pattern.compile("<<(\\w+(:(.+?))?)>>");
     private final static Pattern stmtToken = Pattern.compile("\\?(\\w+[-_\\d\\w]*)|\\{(\\w+[-_\\d\\w.]*)}");
-    private final static String UNSET_VALUE = "UNSET-VALUE";
     private final String stmt;
     private final String name;
-
-    private CQLStatementDef deprecatedDef; // deprecated, to be removed
 
     public void setBindings(Map<String, String> bindings) {
         this.bindings = bindings;
@@ -24,18 +21,9 @@ public class CQLStatementDefParser {
 
     private Map<String, String> bindings;
 
-    public CQLStatementDef getDeprecatedDef() {
-        return deprecatedDef;
-    }
-
-    public void setDeprecatedDef(CQLStatementDef deprecatedDef) {
-        this.deprecatedDef = deprecatedDef;
-    }
-
     public CQLStatementDefParser(String name, String stmt) {
         this.stmt = stmt;
         this.name = name;
-        this.bindings = bindings;
     }
 
     public Map<String,String> getBindings() {
@@ -122,6 +110,7 @@ public class CQLStatementDefParser {
             this.missingGenerators = missingGenerators;
             this.missingAnchors = missingAnchors;
             this.statement = stmt;
+            this.bindings = bindings;
             this.name = name;
         }
 
--- a/driver-cql-shaded/src/test/java/io/nosqlbench/generators/cql/lang/ParserForCqlTest.java
+++ b/driver-cql-shaded/src/test/java/io/nosqlbench/generators/cql/lang/ParserForCqlTest.java
@@ -9,6 +9,7 @@ import java.nio.file.*;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.stream.Stream;
 
 public class ParserForCqlTest {
 
@@ -37,15 +38,14 @@ public class ParserForCqlTest {
                 URL url = resources.nextElement();
                 System.out.println("url=" + url.toExternalForm());
                 Path path = Paths.get(url.toURI());
-                Files.walk(path, FileVisitOption.FOLLOW_LINKS)
-                    .filter(p -> !Files.isDirectory(p, LinkOption.NOFOLLOW_LINKS))
-                    .forEach(subpaths::add);
+                try (Stream<Path> fileStream = Files.walk(path, FileVisitOption.FOLLOW_LINKS)) {
+                    fileStream.filter(p -> !Files.isDirectory(p, LinkOption.NOFOLLOW_LINKS))
+                            .forEach(subpaths::add);
+                }
             }
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
         return subpaths;
-
     }
-
 }
--- a/driver-cqlverify/src/main/java/io/nosqlbench/activitytype/cqlverify/CqlVerifyActivity.java
+++ b/driver-cqlverify/src/main/java/io/nosqlbench/activitytype/cqlverify/CqlVerifyActivity.java
@@ -3,6 +3,7 @@ package io.nosqlbench.activitytype.cqlverify;
 import io.nosqlbench.activitytype.cql.core.CqlActivity;
 import io.nosqlbench.activitytype.cql.statements.rsoperators.AssertSingleRowResultSet;
 import io.nosqlbench.engine.api.activityimpl.ActivityDef;
+import io.nosqlbench.engine.api.activityimpl.ParameterMap;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -22,27 +23,26 @@ public class CqlVerifyActivity extends CqlActivity {
     @Override
     public synchronized void initActivity() {
 
-        if (!super.getActivityDef().getParams().contains("verify") &&
-            !super.getActivityDef().getParams().contains("verify-fields")) {
+        ParameterMap activityParams = super.getActivityDef().getParams();
+        if (!activityParams.containsKey("verify") &&
+            !activityParams.containsKey("verify-fields")) {
             logger.info("Pre-configuring activity param 'verify=*' since none was provided.");
             logger.info("To control this on a per-statement basis, use the verify param.");
-            super.getActivityDef().getParams().put("verify", "*");
+            activityParams.put("verify", "*");
         }
 
-        if (!super.getActivityDef().getParams().contains("compare")) {
-            super.getActivityDef().getParams().put("compare", "all");
+        if (!activityParams.containsKey("compare")) {
+            activityParams.put("compare", "all");
             logger.info("Pre-configuring activity param 'compare=all' since none was provided.");
             logger.info("To control this on a per-statement basis, use the compare param.");
         }
 
         super.initActivity();
-
     }
 
     @Override
     public void onActivityDefUpdate(ActivityDef activityDef) {
         super.onActivityDefUpdate(activityDef);
         addResultSetCycleOperator(new AssertSingleRowResultSet());
-
     }
 }
--- a/driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/PulsarActivity.java
+++ b/driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/PulsarActivity.java
@@ -197,6 +197,6 @@ public class PulsarActivity extends SimpleActivity implements ActivityDefObserve
     }
 
     public void asyncOperationFailed(Throwable ex) {
-        this.asyncOperationFailure = asyncOperationFailure;
+        this.asyncOperationFailure = ex;
     }
 }
--- a/engine-api/src/main/java/io/nosqlbench/engine/api/activityapi/cyclelog/buffers/results_rle/CycleResultsRLEBufferReadable.java
+++ b/engine-api/src/main/java/io/nosqlbench/engine/api/activityapi/cyclelog/buffers/results_rle/CycleResultsRLEBufferReadable.java
@@ -104,11 +104,8 @@ public class CycleResultsRLEBufferReadable implements CycleResultSegmentsReadabl
 
         @Override
         public CycleResultsSegment next() {
-            if (next==null) {
-                hasNext();
-                if (next==null) {
-                    throw new RuntimeException("Possible call to next() without calling hasNext(). There was no remaining unfiltered data.");
-                }
+            if (next == null && !hasNext()) {
+                throw new RuntimeException("Call to next() but there was no remaining unfiltered data.");
             }
             CycleResultsSegment wasNext = this.next;
             next=null;
--- a/engine-api/src/main/java/io/nosqlbench/engine/api/activityimpl/SimpleActivity.java
+++ b/engine-api/src/main/java/io/nosqlbench/engine/api/activityimpl/SimpleActivity.java
@@ -255,7 +255,7 @@ public class SimpleActivity implements Activity, ProgressCapable {
 
     @Override
     public void setPhaseLimiter(RateLimiter rateLimiter) {
-        this.phaseLimiter = phaseLimiter;
+        this.phaseLimiter = rateLimiter;
     }
 
     @Override
--- a/engine-api/src/main/java/io/nosqlbench/engine/api/util/Unit.java
+++ b/engine-api/src/main/java/io/nosqlbench/engine/api/util/Unit.java
@@ -265,17 +265,13 @@ public class Unit {
         TB("TB", "terabyte", bytesPerGB * 1000),
         PB("PB", "petabyte", bytesPerGB * 1000000),
         EB("EB", "exabyte", bytesPerGB * bytesPerGB),
-        ZB("ZB", "zettabyte", bytesPerGB * bytesPerGB * 1000),
-        YB("YB", "yottabyte", bytesPerGB * bytesPerGB * 1000000),
 
         KIB("KiB", "kibibyte", 1024),
         MIB("MiB", "mebibyte", 1024 * 1024),
         GIB("GiB", "gibibyte", BytesPerGiB),
         TIB("TiB", "tebibyte", BytesPerGiB * 1024),
         PIB("PIB", "pebibyte", BytesPerGiB * 1024 * 1024),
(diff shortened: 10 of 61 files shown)


docsys/src/main/java/io/nosqlbench/docsys/endpoints/DocsysMarkdownEndpoint.java
```diff
@@ -180,12 +180,6 @@ private void init(boolean reload) {
     }
 
     private void enable(Set<String> enabled) {
-
-        Set<String> toEnable = new HashSet<>();
-        if (this.enables !=null) {
-            toEnable.addAll(this.enables);
-        }
```
XN137, line 187: unused and the enable method is always called with `this.enables` it seems

driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/datamappers/functions/to_daterange/DateRangeFunc.java
```diff
@@ -75,13 +75,12 @@ public DateRangeFunc(String precision, Function<Long,Long> lower, LongUnaryOpera
...
     }
 
-
     @Override
     public DateRange apply(long value) {
         Date lowerDate = new Date(lower.applyAsLong(value));
         DateRange.DateRangeBound lower = DateRange.DateRangeBound.lowerBound(lowerDate,precision);
         Date upperDate = new Date(upper.applyAsLong(value));
-        DateRange.DateRangeBound upper = DateRange.DateRangeBound.upperBound(lowerDate,precision);
+        DateRange.DateRangeBound upper = DateRange.DateRangeBound.upperBound(upperDate,precision);
```
XN137, line 83: did not work before and might change existing workloads?

driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/statements/core/CQLStatementDefParser.java
```diff
@@ -122,6 +110,7 @@ public ParseResult(String stmt, String name, Map<String,String> bindings, Set<St
             this.missingGenerators = missingGenerators;
             this.missingAnchors = missingAnchors;
             this.statement = stmt;
+            this.bindings = bindings;
```
XN137, line 113: since before it was always null, this probably has no impact in `getBindings()`

driver-cqlverify/src/main/java/io/nosqlbench/activitytype/cqlverify/CqlVerifyActivity.java
```diff
@@ -22,27 +23,26 @@ public CqlVerifyActivity(ActivityDef activityDef) {
     @Override
     public synchronized void initActivity() {
 
-        if (!super.getActivityDef().getParams().contains("verify") &&
-            !super.getActivityDef().getParams().contains("verify-fields")) {
```
XN137, line 26: `ConcurrentHashMap.contains` calls `containsValue` internally... so this never used to work.
what is the impact of now properly setting these param defaults?

driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/PulsarActivity.java
```diff
@@ -197,6 +197,6 @@ public void failOnAsyncOperationFailure() {
     }
 
     public void asyncOperationFailed(Throwable ex) {
-        this.asyncOperationFailure = asyncOperationFailure;
+        this.asyncOperationFailure = ex;
```
XN137, line 200: before this fix `failOnAsyncOperationFailure()` could never throw...

engine-api/src/main/java/io/nosqlbench/engine/api/activityapi/cyclelog/buffers/results_rle/CycleResultsRLEBufferReadable.java
```diff
@@ -104,11 +104,8 @@ public boolean hasNext() {
 
         @Override
         public CycleResultsSegment next() {
-            if (next==null) {
-                hasNext();
-                if (next==null) {
-                    throw new RuntimeException("Possible call to next() without calling hasNext(). There was no remaining unfiltered data.");
-                }
+            if (next == null && !hasNext()) {
```
XN137, line 107: compare with `hasNext` impl/return value above.

engine-api/src/main/java/io/nosqlbench/engine/api/activityimpl/SimpleActivity.java
```diff
@@ -255,7 +255,7 @@ public Timer getResultTimer() {
 
     @Override
     public void setPhaseLimiter(RateLimiter rateLimiter) {
-        this.phaseLimiter = phaseLimiter;
+        this.phaseLimiter = rateLimiter;
```
XN137, line 258: `setPhaseLimiter` doesnt seem to be used anywhere but this was a no-op

engine-api/src/main/java/io/nosqlbench/engine/api/util/Unit.java
```diff
@@ -265,17 +265,13 @@ public double getMultiplier() {
...
-        YB("YB", "yottabyte", bytesPerGB * bytesPerGB * 1000000),
 
         KIB("KiB", "kibibyte", 1024),
         MIB("MiB", "mebibyte", 1024 * 1024),
         GIB("GiB", "gibibyte", BytesPerGiB),
         TIB("TiB", "tebibyte", BytesPerGiB * 1024),
         PIB("PIB", "pebibyte", BytesPerGiB * 1024 * 1024),
-        EIB("EiB", "exbibyte", BytesPerGiB * BytesPerGiB),
-        ZIB("ZiB", "zebibyte", BytesPerGiB * BytesPerGiB * 1024),
-        YIB("YiB", "yobibyte", BytesPerGiB * BytesPerGiB * 1024 * 1024);
```
XN137, line 278: these compile time constants overflowed...
i thought it may be better to remove them (and let users of it fail hard) instead of silenting calculating with invalid values?

engine-api/src/test/java/io/nosqlbench/engine/api/util/TagFilterTest.java
```diff
@@ -110,7 +106,8 @@ public void testMatchingKeyAndValueRegexDoesMatch() {
         assertThat(tfRight.matches(itemtags).matched()).isTrue();
     }
 
-    public void testMatchingDetails() {
+    // TODO: enable and fix test
```
XN137, line 109: i wasnt sure how this is supposed to work, enabling the test caused it to fail

engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/ActivityExecutor.java
```diff
@@ -248,17 +247,14 @@ public boolean finishAndShutdownExecutor(int secondsToWait) {
...
 
-
     /**
      * Listens for changes to parameter maps, maps them to the activity instance, and notifies all eligible listeners of
      * changes.
      */
     @Override
     public synchronized void handleParameterMapUpdate(ParameterMap parameterMap) {
 
-        if (activity instanceof ActivityDefObserver) {
```
XN137, line 259: always true due to type hierarchy... so it was basically a null check but activiy is used below as if it can never be null

engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/ActivityExecutor.java
```diff
@@ -423,8 +419,6 @@ private void awaitActivityAndMotorStateAlignment() {
      * @return true, if the desired SlotState was detected
      */
     private boolean awaitMotorState(Motor m, int waitTime, int pollTime, RunState... desiredRunStates) {
-        Set<RunState> desiredStates = new HashSet<>(Arrays.asList(desiredRunStates));
```
XN137, line 426: the code in the while loop below does not make a lot of sense (populate a map and then remove all entries?).
yet i did not see a way how this set could play a role in making it sane logic.

engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/ScenarioController.java
```diff
@@ -383,9 +383,8 @@ public void forceStopScenario(int waitTimeMillis, boolean rethrow) {
      * @param waitTimeMillis The time to wait, usually set very high
      * @return true, if all activities completed before the timer expired, false otherwise
      */
-    public boolean awaitCompletion(int waitTimeMillis) {
+    public boolean awaitCompletion(long waitTimeMillis) {
```
XN137, line 386: this goes into `Thread.join(long millis)` and it was called with a compile time constant that overflowed the `int` type

engine-core/src/main/java/io/nosqlbench/engine/core/script/Scenario.java
```diff
@@ -311,7 +311,7 @@ public void runScenario() {
                 endedAtMillis = System.currentTimeMillis();
             }
         }
-        int awaitCompletionTime = 86400 * 365 * 1000;
+        long awaitCompletionTime = 86400 * 365 * 1000L;
```
XN137, line 314: compile time constant that overflowed the `int` type:
```
   System.out.println(86400 * 365 * 1000);
   System.out.println(86400 * 365 * 1000L);
```
outputs:
```
1471228928
31536000000
```

engine-core/src/main/java/io/nosqlbench/engine/core/script/ScenariosExecutor.java
```diff
@@ -113,7 +113,7 @@ public ScenariosResults awaitAllResults(long timeout, long updateInterval) {
         }
         Map<Scenario, ScenarioResult> scenarioResultMap = new LinkedHashMap<>();
         getAsyncResultStatus()
-                .entrySet().forEach(es -> scenarioResultMap.put(es.getKey(), es.getValue().orElseGet(null)));
+                .entrySet().forEach(es -> scenarioResultMap.put(es.getKey(), es.getValue().orElse(null)));
```
XN137, line 116: `orElseGet` takes a `Supplier` whic cant be null... this was probably never hit since we would have seen a NPE otherwise

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_long/HostHash.java
```diff
@@ -60,10 +60,10 @@ private static long computeHostHash() {
                 distinctNames.add(iface.getHostName());
             });
             List<String> nameList = new ArrayList<>(distinctNames);
-            nameList.sort(String::compareTo);
+            Collections.sort(nameList);
             Murmur3F m3f = new Murmur3F(0);
             m3f.reset();
-            distinctNames.forEach(
+            nameList.forEach(
```
XN137, line 66: i assume this was the idea behind having a sorted list i.e. more determinism?
otherwise `nameList` was unused.

virtdata-lib-basics/src/test/java/io/nosqlbench/virtdata/library/basics/shared/stateful/ShowTest.java
```diff
@@ -23,8 +23,6 @@ public void testBasicStateSupport() {
         assertThat(showFoo.apply(2342343L)).isEqualTo("{foo=23}");
         assertThat(showBar.apply(23423L)).isEqualTo("{bar=Bar}");
         new Clear().apply(234);
-        assertThat(showAll.apply("234").isEmpty());
+        assertThat(showAll.apply("234")).isEqualTo("{}");
```
XN137, line 26: is that the expected output?

virtdata-lib-basics/src/test/java/io/nosqlbench/virtdata/library/basics/tests/libraryimpl/BasicDataMappersTest.java
```diff
@@ -52,14 +51,17 @@ public void testToDateInstantiator() throws Exception {
...
         Date d2 = dataMapper.get().get(2);
+        assertThat(d2).isAfter(d1);
     }
 
     @Test
     public void testToDateBucketInstantiator() throws Exception {
         Optional<DataMapper<Date>> dataMapper = VirtData.getOptionalMapper("ToDate(1000,10000)");
         assertThat(dataMapper).isNotNull();
         assertThat(dataMapper.get()).isNotNull();
-        assertThat(dataMapper.get().get(1).after(new Date(1)));
```
XN137, line 62: this returned false but no assertion was ever executed... is the new code below the expected logic?

driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/PulsarActivity.java
eolivelli: good catch !

docsys/src/main/java/io/nosqlbench/docsys/endpoints/DocsysMarkdownEndpoint.java
jshook: This does look like a scoping bug. The behavior isn't active in any code so there is no impact. This can be re-added correctly if/when needed.

driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/datamappers/functions/to_daterange/DateRangeFunc.java
jshook: As it could change existing workloads, maybe we should add a call out in release notes for it. However, I do believe the best thing to do is make it correct.

driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/statements/core/CQLSessionCache.java
```diff
@@ -320,7 +320,7 @@ public void afterChannelInitialized(SocketChannel channel) throws Exception
         Cluster cl = builder.build();
 
         // Apply default idempotence, if set
-        activityDef.getParams().getOptionalBoolean("defaultidempotence").map(
+        activityDef.getParams().getOptionalBoolean("defaultidempotence").ifPresent(
```
jshook, line 323: This one could also impact workloads, but should also be made correct. I'll take notes on these through this review to add to the release notes.

XN137: these changes were just cosmetic in that we use the right method (that doesnt return a value we wont use)... previous code still had the same sideeffect i think.

jshook: I believe that the Java streaming semantics mean that the map doesn't cause the stream to be evaluated. I tested with this:
```Java
System.out.println("numbers:");
Stream.of(1,2,3).map(v->{System.out.println("v:"+v); return v;});
System.out.println("letters:");
Stream.of("a","b","c").map(v->{System.out.println("v:"+v); return v;}).collect(Collectors.toList());

```
and got this output:
```
numbers:
letters:
v:a
v:b
v:c
```

driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/statements/core/CQLStatementDefParser.java
jshook: This is (should be) subsumed in the more general parser classes for statements as well.

driver-cql-shaded/src/test/java/io/nosqlbench/generators/cql/lang/ParserForCqlTest.java
```diff
@@ -37,15 +38,14 @@ public void parseAll() {
...
                 System.out.println("url=" + url.toExternalForm());
                 Path path = Paths.get(url.toURI());
-                Files.walk(path, FileVisitOption.FOLLOW_LINKS)
-                    .filter(p -> !Files.isDirectory(p, LinkOption.NOFOLLOW_LINKS))
-                    .forEach(subpaths::add);
+                try (Stream<Path> fileStream = Files.walk(path, FileVisitOption.FOLLOW_LINKS)) {
+                    fileStream.filter(p -> !Files.isDirectory(p, LinkOption.NOFOLLOW_LINKS))
+                            .forEach(subpaths::add);
+                }
             }
```
jshook, line 45: I'm not super clear on the benefit of this additional try scope.

driver-cqlverify/src/main/java/io/nosqlbench/activitytype/cqlverify/CqlVerifyActivity.java
jshook: The logic would force all verify activities to use params verify=* and compare=all, which is the more conservative setting. It would have been a nuisance, because it wouldn't let a user relax the verify constraints.

engine-api/src/main/java/io/nosqlbench/engine/api/activityimpl/SimpleActivity.java
jshook: Phases were/are being _phased_ out. This was too fine-grained of an abstraction based on from its use in the CQL driver for paging.

engine-api/src/main/java/io/nosqlbench/engine/api/util/Unit.java
jshook: agreed

driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/statements/core/CQLSessionCache.java
XN137: the type here is `Optional` not `Stream` so:
```
    private static Object printStuff(String s)
    {
        System.out.println(s);
        return null;
    }

    public static void main(String[] args)
    {
        Optional.of("foo").map(s -> printStuff(s));
        Optional.<String>empty().map(s -> printStuff(s));
        Optional.of("bar").ifPresent(s -> printStuff(s));
        Optional.<String>empty().ifPresent(s -> printStuff(s));
    }
```
prints:
```
foo
bar
```

driver-cql-shaded/src/test/java/io/nosqlbench/generators/cql/lang/ParserForCqlTest.java
XN137:
```
The returned stream encapsulates one or more DirectoryStreams. If timely disposal of file system resources is required, the try-with-resources construct should be used to ensure that the stream's close method is invoked after the stream operations are completed.
```
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#walk-java.nio.file.Path-java.nio.file.FileVisitOption...-

so making sure resources are cleared up immediately

engine-api/src/test/java/io/nosqlbench/engine/api/util/TagFilterTest.java
jshook: I've updated the test. The flow has changed since the test was originally written with an earlier test framework that didn't require per-method annotations.
The change in flow was to keep diag logs more concise without notifying the user of additional predicate misses once one has already missed.

driver-cql-shaded/src/main/java/io/nosqlbench/activitytype/cql/statements/core/CQLSessionCache.java
jshook: Oh, right!

engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/ActivityExecutor.java
jshook: It looks like I was trying to tackle a concurrency bug with some stack-local state checking, but missed some scoping (and probably made it even less clear in the process). I'm reviewing the logic here and will probably need to add some test to shore up any changes here due to the cross-cutting impact of changes on this class.

jshook: Ah, I see what the logic was doing.
There is obviously a scoping bug, however:
The actual and desired states are both counted, then the actual states which are in fact also the desired states are removed from actual states, thus leaving the remaining states which were not desired. This simply provides a summary view which is good for diagnostics. (Also, the result of this is used for control flow)

I suspect the scoping issue was from earlier code in which the actual states were kept as a copy and mutated locally.

engine-core/src/main/java/io/nosqlbench/engine/core/script/ScenariosExecutor.java
jshook: The async view is needed for web services, but I've never seen an NPE or other issue from this code. But yes, the intent was for a value not a supplier.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_long/HostHash.java
jshook: Correct, corrected!

virtdata-lib-basics/src/test/java/io/nosqlbench/virtdata/library/basics/shared/stateful/ShowTest.java
jshook: It is now. This was always a diagnostic function, and using json _object_ formatting is a simplification I think.

driver-cqlverify/src/main/java/io/nosqlbench/activitytype/cqlverify/CqlVerifyActivity.java
XN137: not sure i understand... my suggested changes here already change these to use `containsKey` instead

virtdata-lib-basics/src/test/java/io/nosqlbench/virtdata/library/basics/tests/libraryimpl/BasicDataMappersTest.java
jshook: The semantics of this function are a bit muddled. I'm adding unit tests and appropriate documentation. The second param is indicated as a "repeat" factor, but It looks sus. I'll clean it up with unit tests.

jshook (approved):
Lots of good improvements in this PR. I'm happy to merge it and handle any fixups locally before the next release.

driver-cqlverify/src/main/java/io/nosqlbench/activitytype/cqlverify/CqlVerifyActivity.java
jshook: Yep, I glossed over that detail!

driver-cql-shaded/src/test/java/io/nosqlbench/generators/cql/lang/ParserForCqlTest.java
jshook: Yes, somehow I missed the twist on try-with-resources. I see it now.

PR #337, opened 2021-06-18
unify antlr4 version

XN137:
sorry for missing this earlier in https://github.com/nosqlbench/nosqlbench/pull/335

we further unify this by only defining the antlr4 version once.

also i noticed that the tests on main are failing, which is fixed by the 2nd commit.

### DIFF
--- a/mvn-defaults/pom.xml
+++ b/mvn-defaults/pom.xml
@@ -13,6 +13,7 @@
 
     <!-- properties for package versions -->
     <ascii.data.version>1.2.0</ascii.data.version>
+    <antlr4.version>4.9.2</antlr4.version>
     <commons.codec.version>1.14</commons.codec.version>
     <commons.compress.version>1.20</commons.compress.version>
     <commons.csv.version>1.8</commons.csv.version>
@@ -270,7 +271,7 @@
       <dependency>
         <groupId>org.antlr</groupId>
         <artifactId>antlr4-runtime</artifactId>
-        <version>4.9.2</version>
+        <version>${antlr4.version}</version>
       </dependency>
 
       <dependency>
@@ -547,7 +548,7 @@
         <plugin>
           <groupId>org.antlr</groupId>
           <artifactId>antlr4-maven-plugin</artifactId>
-          <version>4.9.2</version>
+          <version>${antlr4.version}</version>
         </plugin>
         <plugin>
           <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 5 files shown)


virtdata-userlibs/src/test/java/io/virtdata/IntegratedComposerLogicTest.java
```diff
@@ -136,10 +136,8 @@ public void sanityCheckFunctionCasting() {
         assertThat(ClassUtils.isAssignable(c1, c2)).isTrue();
 
         Class<?> c3 = Identity.class;
-        Class<?> c4 = LongFunction.class;
+        Class<?> c4 = LongUnaryOperator.class;
         assertThat(ClassUtils.isAssignable(c3, c4)).isTrue();
-        LongUnaryOperator f;
-        f = new Identity();
```
XN137, line 142: i could have sworn that i ran the test when i added the missing `isTrue()` but since the tests are failing on the main branch right now, i was probably imagening things.

i went back to a commit before any of my recent changes and this never returned `true` as is... the code below already showed that `Identity` was only ever assignable to `LongUnaryOperator` not `LongFunction`.

XN137: not sure what is going on but on the main branch sometimes `mvn clean package` passes as well... where as it always fail when run via intellij... at least with this change this code looks more correct in this test? :sweat_smile:

jshook: I think this is working a little differently depending on the generation of JVM. There have been some subtle changes since the functional types were added, it seems.
I'm just fixing these up to current observed behavior, as it doesn't seem critical path to the other type finesses that are built-in to the current function composer logic.

mvn-defaults/pom.xml
```diff
@@ -547,7 +548,7 @@
         <plugin>
           <groupId>org.antlr</groupId>
           <artifactId>antlr4-maven-plugin</artifactId>
-          <version>4.9.2</version>
```
jshook, line 550: I have a change in my post-merge fixups for these locally. I'll merge or close this PR soon.

jshook:
@XN137 I propose we close this PR, since I fixed up the version alignment. If we still want to centralize the version, we can do another PR?

XN137:
> I propose we close this PR, since I fixed up the version alignment. If we still want to centralize the version, we can do another PR?

i force pushed the branch to not include the 2nd commit anymore, so this PR should be good to review again.

PR #339, opened 2021-06-28
Adds support in NBIO for S3 URI's

szimmer1:
UX notes:

- Credentials and region must be set either in environment variables or an ~/.aws/... file ([doc](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-credentials.html)). Wasn't sure where to document this.
- Warning log is emitted unless there are both .yml and .yaml extension files available in the given path. I don't know how to avoid this due to the way ResolverForS3 is called: it doesn't have a global view of all paths requested by the CLI, so it can't determine if "at least one file was found". For example, only the .yaml extension file exists:
    ```
    ./nb run type=stdout workload=s3://my-bucket/my-prefix/s3-postgres-basic tags=phase:schema
       6156 WARN  [scenarios:001] ResolverForS3 Failed to access S3 object at 's3://my-bucket/my-prefix/s3-postgres-basic.yml': The specified key does not exist. (Service: Amazon S3; Status Code: 404; Error Code: NoSuchKey; Request ID: ZYQV5ZCC9Z36RYG4; S3 Extended Request ID: 7TQ4fcG1RJfBPlwz6FAgnugc6bR3o6hDTAAGlKa91VHh39pI60xD6AQi+X1AyP9U5I7f24Jw4xQ=; Proxy: null)
    CREATE TABLE IF NOT EXISTS "account2" (
      uuid UUID PRIMARY KEY,
      amount INTEGER,
      amount_unit VARCHAR(64),
      updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
      created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
      filler TEXT
    );
    CREATE INDEX IF NOT EXISTS amount_idx on "account2" (amount);
    CREATE INDEX IF NOT EXISTS updated_at_idx on "account2" (updated_at);
    ```

New dependencies added:

- com.amazonaws:aws-java-sdk-s3:1.12.12
- org.mockito:mockito-core:3.3.3

Implementation notes:

- Refactored common parts of existing URLContent and new URIContent to parent class StreamContent

### DIFF
--- a/nb-api/src/main/java/io/nosqlbench/nb/api/content/StreamContent.java
+++ b/nb-api/src/main/java/io/nosqlbench/nb/api/content/StreamContent.java
@@ -1,0 +1,53 @@
+package io.nosqlbench.nb.api.content;
+
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.CharBuffer;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * StreamContent is meant for short-lived use as an efficient way to
+ * find a read URL content. If a caller has already acquired an
+ * input stream, it can be passed to the stream content holder
+ * to avoid double fetch or other unintuitive and inefficient
+ * behavior.
+ *
+ * Subclasses must implement getURI()
+ */
+public abstract class StreamContent<T> implements Content<T> {
+    private final T location;
+    private final InputStream inputStream;
+
+    public StreamContent(T location, InputStream inputStream) {
+        this.location = location;
+        this.inputStream = inputStream;
+    }
+
+    @Override
+    public T getLocation() {
+        return location;
+    }
+
+    @Override
+    public CharBuffer getCharBuffer() {
+        InputStreamReader inputStreamReader = new InputStreamReader(inputStream, StandardCharsets.UTF_8);
+        BufferedReader bufferedReader = new BufferedReader(inputStreamReader);
+        Stream<String> lines = bufferedReader.lines();
+        String buffdata = lines.map(l -> l+"\n").collect(Collectors.joining());
+        return CharBuffer.wrap(buffdata);
+    }
+
+    @Override
+    public InputStream getInputStream() {
+        return inputStream;
+    }
+
+    @Override
+    public Path asPath() {
+        return null;
+    }
+}
--- a/nb-api/src/main/java/io/nosqlbench/nb/api/content/URIContent.java
+++ b/nb-api/src/main/java/io/nosqlbench/nb/api/content/URIContent.java
@@ -1,0 +1,35 @@
+package io.nosqlbench.nb.api.content;
+
+import java.io.InputStream;
+import java.net.URI;
+import java.util.Objects;
+
+public class URIContent extends StreamContent<URI> {
+    private final URI uri;
+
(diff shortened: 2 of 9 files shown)


nb-api/src/main/java/io/nosqlbench/nb/api/content/StreamContent.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public T getLocation() {
+        return location;
+    }
+
+    @Override
+    public CharBuffer getCharBuffer() {
+        InputStreamReader inputStreamReader = new InputStreamReader(inputStream, StandardCharsets.UTF_8);
```
jshook, line 37: This is missing the lazy-init and cache behavior of the previous implementation. This means that that readers will exhaust the stream when read a second time. Please cache and return the buffer instead. These objects will be disposed once out of scope anyway.

nb-api/src/test/java/io/nosqlbench/nb/api/pathutil/ResolverForURLTest.java
```diff
@@ -1,19 +1,38 @@
...
+            .thenReturn(new ByteArrayInputStream(text.getBytes()));
+        ResolverForS3 r = new ResolverForS3(mockS3Access);
+        List<Content<?>> content = r.resolve(s3URI);
+        assertThat(content).isNotNull();
+        assertThat(content).isNotEmpty();
+        Content<?> c = content.get(0);
+        Object location = c.getLocation();
+        assertThat(location).isInstanceOf(URI.class);
+        assertThat(location.toString()).isEqualTo(s3URI);
+        String actualText = c.getCharBuffer().toString();
```
jshook, line 32: Please add coverage for double read on the char buffer to illustrate.

jshook (changes requested):

nb-api/src/main/java/io/nosqlbench/nb/api/content/URIContent.java
```diff
@@ -0,0 +1,35 @@
+package io.nosqlbench.nb.api.content;
+
+import java.io.InputStream;
+import java.net.URI;
+import java.util.Objects;
+
+public class URIContent extends StreamContent<URI> {
```
jshook, line 7: After some research, it appears that we can simply implement a https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/spi/URLStreamHandlerProvider.html and then have only URLs here. Are you interested in prototyping this?

jshook: In other words, we can keep closer to the meaning of URLs and also avoid having two version of this concept here. It also would be a good illustration for how to add these for other protocols in the Java-prescribed way.

szimmer1: Will absolutely take a look into the URLStreamHandlerProvider

jshook:
@szimmer1 I'm going to take a swing at adding the S3 URL handler as we discussed above.

jshook:
@szimmer1 The latest release has S3 support in the form of a URL Handler as well as a new scripting plugin for S3.
This means that NBIO should automatically work for S3 resources as long as you have your ~/.aws/ files configured OR you add it to the URL as digest authentication. (This is extracted to support S3 client auth config).

Do you mind taking a look and testing it out with NBIO? If you are happy with the functionality of it, we can update the docs and retire this PR.

PR #342, opened 2021-07-06
Refactor error format for error handling in JDBCActivity

szimmer1:
Add examples for default error handler specs for cockroabdb-basic and postgres-basic workloads.
Fixes a retry bug in JDBCAction where previously max retries wouldn't throw an error.


jshook (approved):
Looks good to me

PR #343, opened 2021-07-20
Update Pulsar API to 2.8.0

yabinmeng:


driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/PulsarAdminTenantOp.java
```diff
@@ -45,7 +45,8 @@ public void run() {
 
         // Admin API - create tenants and namespaces
         if (!adminDelOp) {
-            TenantInfo tenantInfo = new TenantInfo();
+
+            TenantInfoImpl tenantInfo = TenantInfoImpl.builder().build();
```
eolivelli, line 49: This should be TenantInfo.builder() probably and we should not use the Impl type

eolivelli (commented):
Awesome work

PR #348, opened 2021-08-12
This should fix 404's during the main phase of the documents API test

EricBorczuk:
Before this change, using default params, the rampup phase of the Docs API workload was performing 10^7 inserts into a table. Then, the main phase runs, but its reads (and write!) are selecting values from 1 to 10^9. What this means is that there was only a 1% chance (10^7 / 10^9 as a percent) that a key that was generated for read actually existed in the data created by the ramp-up.  As the workflow ran further, since writes and reads are done in parallel, new data was written to keys that weren't written by the ramp-up, but in reality we want the ramp-up to populate all the data from the beginning, for reads.

### DIFF
--- a/driver-http/src/main/resources/activities/documents-api/http-docsapi-keyvalue.yaml
+++ b/driver-http/src/main/resources/activities/documents-api/http-docsapi-keyvalue.yaml
@@ -23,9 +23,9 @@ bindings:
   # http request id
   request_id: ToHashedUUID(); ToString();
 
-  seq_key: Mod(<<keycount:1000000000>>); ToString() -> String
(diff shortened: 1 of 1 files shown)


jshook (approved):
Thanks! Looks good.

jshook:
For more impacting changes, like the distribution, these should be added to the baselinesv2 directory.

ivansenic:
Can we go with #349 instead of this one, as it contains further fixes. We can agree if we wanna port these changes on top, but let's discuss.

EricBorczuk:
Closing this in favor of https://github.com/nosqlbench/nosqlbench/pull/349 which does the same thing but has other changes too

PR #350, opened 2021-08-13
Fix a few keyvalue test cases

EricBorczuk:
Relates to https://github.com/nosqlbench/nosqlbench/pull/349

While https://github.com/nosqlbench/nosqlbench/pull/349 fixes the docs-api-keyvalue test, this does a similar set of fixes for http-rest-keyvalue, http-graphql-cql-keyvalue, and http-graphql-keyvalue.

The tabular/timeseries tests have their own special way to match keys with the cycle number, and upon review it looks like they are behaving properly.

Also note that the graphql-related tests do NOT accept 4XX as a valid response, this is because graphql is not ever meant to return 4XX for valid requests.

### DIFF
--- a/driver-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
+++ b/driver-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
@@ -7,9 +7,9 @@ description: |
 
 scenarios:
   default:
-    - run driver=cql tags==phase:schema threads==1 cycles==UNDEF
-    - run driver=http tags==phase:rampup cycles===TEMPLATE(rampup-cycles,10000000) threads=auto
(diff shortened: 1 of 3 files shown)


jshook (approved):
LGTM

ivansenic:
Yep let's merge this one and the #349

ivansenic:
@jshook @EricBorczuk I just noticed that the changes in the PR missed to update the driver type in the `schema` phase for the default scenario to `http`, as we are having HTTP calls and not the CQL ones. Eric, can you please revisit and fix this in separate PR and can you actually make sure that the default scenario is running fine. I guess that the current one would fail due to this error, maybe we missed something else in addition:

`nb ./http-rest-keyvalue.yaml auth_token=.. rampup-cycles=.. main-cycles=..`

PR #359, opened 2021-09-22
NB Pulsar driver enhancement - Milestone 11

yabinmeng:
(https://github.com/nosqlbench/nosqlbench/milestone/11)

- add support for message properties
- add support for end-to-end latency measurement (latency histogram)
- add support for message out-of-order and loss detect
- add async api for consumer
- split single-topic consumer and multi-topic consumer
- SSL/TLS bug fix
- Code cleanup

### DIFF
--- a/driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/PulsarConsumerOp.java
+++ b/driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/PulsarConsumerOp.java
@@ -3,6 +3,7 @@ package io.nosqlbench.driver.pulsar.ops;
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Timer;
+import io.nosqlbench.driver.pulsar.PulsarActivity;
 import io.nosqlbench.driver.pulsar.util.AvroUtil;
 import io.nosqlbench.driver.pulsar.util.PulsarActivityUtil;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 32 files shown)


eolivelli (approved):
overall looks good to me
it is a big patch so I am not sure I reviewed it too carefully.
btw it looks like it is compatible with the previous version (we have many fallout haxx tests that depend on this)

jshook (approved):
Quite a few enhancements.  Looks like a solid set of changes!
I'm going to merge and release this today.

Only one comment: "lower level" is not clear in the docs It might good to find a more explicit wording.

driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/PulsarConsumerOp.java
```diff
@@ -11,106 +12,260 @@
...
+                    else {
+                        // abnormal case: out of ordering
+                        if (curMsgSeqId < prevMsgSeqid) {
+                            throw new RuntimeException("Detected message ordering is not guaranteed. Older messages are received earlier!");
+                        }
+                        // abnormal case: message loss
+                        else if ( (curMsgSeqId - prevMsgSeqid) > 1 ) {
+                            throw new RuntimeException("Detected message sequence id gap. Some published messages are not received!");
+                        }
+                    }
```
lhotari, line 156: It would be useful to have counters for detected message loss / out-of-order delivery. There is also the case of duplicate message delivery which is expected behavior unless de-duplication feature of Pulsar is used.

duplicates can happen in producing as explained in https://jack-vanlightly.com/blog/2018/10/21/how-to-not-lose-messages-on-an-apache-pulsar-cluster :
> Finally, regarding duplicates. The Pulsar client will automatically retry messages on failure up to the send timeout. This means that duplicates can be created when a message successfully gets written to BookKeeper but the Pulsar broker dies or fails before it can send an ack to the client.

If we'd like to support a way where the checks are made in a streaming fashion, I guess the algorithm would have to support streaming by using a window (or sliding window) for determining whether a message is lost or delivered out of order. 
Duplicate detection probably doesn't require a windowing algorithm, since checking for a previous sequence id would be sufficient for duplicate detection?

@yabinmeng What are your thoughts in improving this area?

yabinmeng: @lhotari here are my 2 cents. 

1). Regarding counter metrics for message loss / out-of-order delivery, do we really need it? My thinking here is as long as we detect one such occurrence, that's a critical error and we should stop. The criticality of having one such occurrence is as big as have many and there is no point to count how many such occurrences are
2). Regarding duplication, would it be considered an error for the application (if de-duplication is not enabled)? From Pulsar's perspective, this is a normal behavior in order to avoid message loss. I guess what we should add in NB Pulsar driver is if  de-duplication is enabled, then we should catch the case if the consumer receives duplicate messages.
3). Regarding your "streaming" comment with windowing, I don't fully understand and let's chat.

jshook:
@all-contributors please add @yabinmeng for drivers, bug fixes, API testing, and issues

PR #362, opened 2021-10-02
Ignore message out-of-order and duplicate check for Shared and Key_Shared

yabinmeng:
TODO: such checks can be applied to Key_Shared subscription. But it needs to be applied at per-key basis. Right now this is function is missing; add it in future release


jshook (approved):
LGTM, merging. I'll kick a release build today.

PR #364, opened 2021-10-06
Fix NullPointerException to check namespce/topic dedup status 

yabinmeng:


driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/ReadyPulsarOp.java
```diff
@@ -461,7 +461,7 @@ else if (StringUtils.equalsIgnoreCase(stmtOpType, PulsarActivityUtil.OP_TYPES.E2
             try {
                 nsMsgDedup = pulsarAdmin.namespaces().getDeduplicationStatus(namespace);
```
lhotari, line 462: I think there should be code like this to fix the NPE:
```
Boolean dedupStatus = pulsarAdmin.namespaces().getDeduplicationStatus(namespace);
nsMsgDedup = dedupStatus != null ? dedupStatus : false;
```

driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/ReadyPulsarOp.java
```diff
@@ -471,7 +471,7 @@ else if (StringUtils.equalsIgnoreCase(stmtOpType, PulsarActivityUtil.OP_TYPES.E2
             try {
                 topicMsgDedup = pulsarAdmin.topics().getDeduplicationStatus(topic);
```
lhotari, line 472:
```
Boolean dedupStatus = pulsarAdmin.topics().getDeduplicationStatus(namespace);
topicMsgDedup = dedupStatus != null ? dedupStatus : false;
```

lhotari (changes requested):

driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/ReadyPulsarOp.java
jshook: We shouldn't allow NPEs to be swallowed this way. Instead, get the higher level object first and check for null.
It's just too risky to swallow exceptions like this. It's not entirely clear that an NPE is the only way this could error, and also that that is always ok.

jshook (changes requested):
We shouldn't let NPEs slip through in such a generic way. They are a frequent source of user pain. Please add more defensive code around the actual calls and handle null values explicitly.

yabinmeng:
@lhotari @jshook thanks for the comments; addressed

jshook (changes requested):
@yabinmeng Just one more detail, then we're good for merge.

yabinmeng:
Not sure how an explicit conditional logic can be done here without throwing the exception out which I definitely don't want to do.

PR #369, opened 2021-10-19
Make ratelimiter instance thread specific for PulsarActivity

lhotari:
### Motivation

Makes the ratelimiter instance thread specific for the Pulsar driver operations.
use case: producing to individual topics where rate should be limited by topic, for example 500 msgs/s per topic across 100 topic partitions. This can be achieved with this type of scenario when threads=100:

```yaml
bindings:
  partition_num: ThreadNum(); Mod(100); ToString()
params:
  topic_uri: persistent://public/perf/perftest-partition-{partition_num}
  async_api: 'true'
blocks:
  - name: producer
    statements:
      - name: produce
        optype: ec2-msg-proc-send
        async_api: 'true'
        msg_value: "hello world!"
```

### Modifications
- `cyclerate` becomes thread specific when `cyclerate_per_thread` is set to true
- only supported in the Pulsar driver since this is an intermediate solution.


yabinmeng:
Would it be an issue when thread specific rate limit is not aligned with "global" stride rate?

PR #374, opened 2021-11-02
add instrumentation scaffold for payloadRtt

jshook:
This puts in a separate rtt tracking mechanism which may work with different sender and receiver types, since it can rely on a payload field instead of intrinsic elements of the pulsar API. I have left the last piece out -- the extraction and conversion of the send time value.

### DIFF
--- a/driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/PulsarConsumerOp.java
+++ b/driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/PulsarConsumerOp.java
@@ -4,19 +4,21 @@ import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Timer;
 import io.nosqlbench.driver.pulsar.PulsarActivity;
-import io.nosqlbench.driver.pulsar.exception.*;
+import io.nosqlbench.driver.pulsar.exception.PulsarDriverUnexpectedException;
 import io.nosqlbench.driver.pulsar.util.AvroUtil;
 import io.nosqlbench.driver.pulsar.util.PulsarActivityUtil;
-import java.util.function.Function;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.pulsar.client.api.*;
+import org.apache.pulsar.client.api.Consumer;
+import org.apache.pulsar.client.api.Message;
+import org.apache.pulsar.client.api.Schema;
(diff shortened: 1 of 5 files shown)


mfleming:
@jshook this branch works for my use case now. What's needed to get this merged into `main`?

driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/PulsarConsumerOp.java
```diff
@@ -147,10 +148,16 @@ public void run(Runnable timeTracker) {
...
+                    org.apache.avro.Schema avroSchema =
+                            AvroUtil.GetSchema_ApacheAvro(avroDefStr);
+                    org.apache.avro.generic.GenericRecord avroGenericRecord =
+                            AvroUtil.GetGenericRecord_ApacheAvro(avroSchema, message.getData());
+                    if (avroGenericRecord.hasField(payloadRttTrackingField)) {
+                        long extractedSendTime = (Long)avroGenericRecord.get(payloadRttTrackingField);
+                        long delta = System.currentTimeMillis() - extractedSendTime;
+                        payloadRttHistogram.update(delta);
+                    }
                 }
```
jshook, line 161: There is substantial initialization logic here in the case that schema has to be loaded. This should be pushed towards the front of the activity's lifecycle. I'll merge it now so you can use the code, but it should get an "init vs run fast" optimization, as schema realization is likely to be non-trivial.

jshook (commented):
There is one change we should make, but I'm comfortable merging it in.

driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/PulsarConsumerOp.java
eolivelli: Also it would have been better to use a message header instead of putting this field into the payload.
AVRO adds overhead and here we are measuring latency

eolivelli: my understanding is that this feature was already implemented using "e2e_starting_time"

mfleming: This is not the same as `e2e_starting_time` AFAIK because that would require CNDB to extract the embedded start timestamp payload from the CQL INSERT and add it as a property to the pulsar message. `payloadRtt` doesn't require that kind of munging in theory (in practice, I never implemented a generic "map CQL types to AVRO types" method and instead pull out a specific column name but `payloadRtt` is more generic on paper).

PR #378, opened 2021-11-09
Document API search endpoint workloads

EricBorczuk:
This adds two new workloads in the http-driver's resources for testing with the Stargate Documents API.

The tests specifically test the search functionality in the Docs API. The two workloads are for "basic" functionality (i.e. filters that are supported by cassandra, using a single AND or OR), and "advanced" functionality (i.e. filters that are unsupported in various combinations, with more complex boolean logic)

For each of these tests, one can define the following special parameters for the write phase:

`match-ratio` (between 0 and 1): defines how many documents to write that match the filters during the write phase (a value of 0.1 for example would mean that 1 of every 10 documents should match the search filters)

`docpadding` (defaults to 0): defines how large each document should be; adds values to each document to make them larger.

`docscount` (defaults to 10M): defines how many documents should be written during the write phase

And the following special parameters for the read phase:

`fields` (defaults to all): which fields to select

`page-size` (defaults to 3, max 20): how many documents to limit the search to

### DIFF
--- a/driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.md
+++ b/driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.md
@@ -50,18 +50,19 @@ The example JSON looks like:
 ## Filters tested
 
 The advanced search workload can test the following `where` clauses:
-- in: match1 IN [0]
-- not-in: match2 NOT IN ["false"]
-- mem-and: match2 EQ "true" AND match3 NOT EQ false
-- mem-or: match1 LT 1 OR match3 EXISTS
-- complex1: match1 EQ 0 AND (match2 EQ "true" OR match3 EQ false)
-- complex2: (match1 LTE 0 OR match2 EQ "false") AND (match2 EQ "false" OR match3 EQ true)
-- complex3: (match1 LTE 0 AND match2 EQ "true") OR (match2 EQ "false" AND match3 EQ true)
+- in: `match1 IN [0]`
+- not-in: `match2 NOT IN ["false"]`
+- mem-and: `match2 EQ "true" AND match3 NOT EQ false`
+- mem-or: `match1 LT 1 OR match3 EXISTS`
+- complex1: `match1 EQ 0 AND (match2 EQ "true" OR match3 EQ false)`
+- complex2: `(match1 LTE 0 OR match2 EQ "false") AND (match2 EQ "false" OR match3 EQ true)`
+- complex3: `(match1 LTE 0 AND match2 EQ "true") OR (match2 EQ "false" AND match3 EQ true)`
 
 ## Workload Parameters
 
 - `docscount` - the number of documents to write during rampup (default: `10_000_000`)
 - `docpadding` - the number of fields to add to each document; useful for writing larger documents. A value of e.g. `5` would make each document have 20 leaf values, instead of 15. (default: `0`)
 - `match-ratio` - a value between 0 and 1 detailing what ratio of the documents written should match the search parameters. If match-ratio is e.g. `0.1` then approximately one-tenth of the documents will have `match1`, `match2`, and `match3` values that are `0`, `"true"`, and `true`, respectively. (default: `0.01`)
+- `fields` - the URL-encoded value for `fields` that you would send to the Docs API. This restricts the fields returned during benchmarking.
 
 
--- a/driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.md
+++ b/driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.md
@@ -1,0 +1,67 @@
+---
+title: Documents API Search Advanced
+weight: 2
+---
+
+## Description
+
+The Documents API Search Advanced workflow targets Stargate's Documents API using generated JSON documents. Specifically, it looks to benchmark more advanced search cases, using both complex boolean logic and filters that aren't natively supported by the underlying data store.
+By default, the documents used are sharing the same structure and are approximately half a kilobyte in size each:
+
+* Each document has 15 leaf values, with a maximum depth of 3
+* there is at least one `string`, `boolean`, `number` and `null` leaf
+* there is one array with `double` values and one with `string` values
+* there is one empty array and one empty map
+
+The example JSON looks like:
+
+```json
+{
+  "user_id":"56fd76f6-081d-401a-85eb-b1d9e5bba058",
+  "created_on":1476743286,
+  "full_name":"Andrew Daniels",
+  "married":true,
+  "address":{
+    "primary":{
+      "cc":"IO",
+      "city":"Okmulgee"
+    },
+    "secondary":{
+
+    }
+  },
+  "coordinates":[
+    64.65964627052323,
+    -122.35334535072856
+  ],
+  "children":[
+
+  ],
+  "friends":[
+    "3df498b1-9568-4584-96fd-76f6081da01a"
+  ],
+  "debt":null,
+  "match1": 0, // or 1000
+  "match2": "true", // or "false"
+  "match3": true // or false
+}
+```
+
+## Filters tested
+
+The advanced search workload can test the following `where` clauses:
+- in: match1 IN [0]
+- not-in: match2 NOT IN ["false"]
+- mem-and: match2 EQ "true" AND match3 NOT EQ false
+- mem-or: match1 LT 1 OR match3 EXISTS
+- complex1: match1 EQ 0 AND (match2 EQ "true" OR match3 EQ false)
+- complex2: (match1 LTE 0 OR match2 EQ "false") AND (match2 EQ "false" OR match3 EQ true)
+- complex3: (match1 LTE 0 AND match2 EQ "true") OR (match2 EQ "false" AND match3 EQ true)
+
+## Workload Parameters
+
+- `docscount` - the number of documents to write during rampup (default: `10_000_000`)
+- `docpadding` - the number of fields to add to each document; useful for writing larger documents. A value of e.g. `5` would make each document have 20 leaf values, instead of 15. (default: `0`)
+- `match-ratio` - a value between 0 and 1 detailing what ratio of the documents written should match the search parameters. If match-ratio is e.g. `0.1` then approximately one-tenth of the documents will have `match1`, `match2`, and `match3` values that are `0`, `"true"`, and `true`, respectively. (default: `0.01`)
+
+
--- a/driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
+++ b/driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
@@ -16,7 +16,7 @@ description: |
 # complex3: (match1 LTE 0 AND match2 EQ "true") OR (match2 EQ "false" AND match3 EQ true)
 scenarios:
   schema:             run driver=http tags==phase:schema threads==<<threads:1>> cycles==UNDEF
-  rampup-write:       run driver=http tags==phase:rampup-write cycles===TEMPLATE(docscount,10000000) docpadding=TEMPLATE(docpadding,15) match-ratio=TEMPLATE(match-ratio,0.01) threads=<<threads:auto>> errors=timer,warn
+  rampup-write:       run driver=http tags==phase:rampup-write cycles===TEMPLATE(docscount,10000000) docpadding=TEMPLATE(docpadding,0) match-ratio=TEMPLATE(match-ratio,0.01) threads=<<threads:auto>> errors=timer,warn
   rampup-read:        run driver=http tags==phase:rampup-read cycles===TEMPLATE(rampup-cycles, 10000000) page-size=TEMPLATE(page-size,3) fields=TEMPLATE(fields,%5b%5d) threads=<<threads:auto>> errors=timer,warn
   main:               run driver=http tags==phase:main cycles===TEMPLATE(read-cycles,TEMPLATE(docscount,10000000)) page-size=TEMPLATE(page-size,3) fields=TEMPLATE(fields,%5b%5d) threads=<<threads:auto>> errors=timer,warn
   main-in:            run driver=http tags==phase:main,filter:in cycles===TEMPLATE(read-cycles,TEMPLATE(docscount,10000000)) page-size=TEMPLATE(page-size,3) fields=TEMPLATE(fields,%5b%5d) threads=<<threads:auto>> errors=timer,warn
--- a/driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
+++ b/driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
@@ -16,7 +16,7 @@ description: |
 # complex3: (match1 LTE 0 AND match2 EQ "true") OR (match2 EQ "false" AND match3 EQ true)
 scenarios:
   schema:             run driver=http tags==phase:schema threads==<<threads:1>> cycles==UNDEF
-  rampup-write:       run driver=http tags==phase:rampup-write cycles===TEMPLATE(docscount,10000000) docsize=TEMPLATE(docsize,15) match-ratio=TEMPLATE(match-ratio,0.01) threads=<<threads:auto>> errors=timer,warn
+  rampup-write:       run driver=http tags==phase:rampup-write cycles===TEMPLATE(docscount,10000000) docpadding=TEMPLATE(docpadding,15) match-ratio=TEMPLATE(match-ratio,0.01) threads=<<threads:auto>> errors=timer,warn
   rampup-read:        run driver=http tags==phase:rampup-read cycles===TEMPLATE(rampup-cycles, 10000000) page-size=TEMPLATE(page-size,3) fields=TEMPLATE(fields,%5b%5d) threads=<<threads:auto>> errors=timer,warn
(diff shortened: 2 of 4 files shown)


driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
```diff
@@ -0,0 +1,315 @@
...
+  created_on: Uniform(1262304000,1577836800) -> long
+  full_name: FullNames()
+  married: ModuloToBoolean()
+  city: Cities()
+  country_code: CountryCodes()
+  lat: Uniform(-180d, 180d)
+  lng: Hash() -> long; Uniform(-180d, 180d)
+  friend_id: Add(-1); ToHashedUUID(); ToString() -> String
+
+  match1: Expr('if (cycle % (int)(1 / <<match-ratio>>) == 0) { return 0; } return (int)(Math.random() * 10000 + 1000);') -> long
```
EricBorczuk, line 53: What I've done here on lines 53-56 is pretty hacky - to explain the `match` values, I want them to equal a "matching" value only when the cycle is mod the match ratio's inverse (e.g if the match ratio is `0.5` then every 2nd cycle should be a match) - is there a way to do this without `Expr` and MVEL?

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
```diff
@@ -0,0 +1,315 @@
...
+  match2: Expr('if (cycle % (int)(1 / <<match-ratio>>) == 0) { return "true" } return "false";') -> String
+  match3: Expr('if (cycle % (int)(1 / <<match-ratio>>) == 0) { return "true" } return "false";') -> String
+  
+  additional_fields: Expr('str="";for (int i=0; i < TEMPLATE(docsize,15) - 15; i++) { if (str == "") { str = "," }; str+="\"value"+i+"\":0"; if (i < TEMPLATE(docsize,15) - 16) {str += ","}} return str;') -> String
```
EricBorczuk, line 57: This is also pretty hacky - I want additional_fields to be a string that is tacked onto the JSON body below, based on the `docsize` parameter. There are by default 15 fields defined, so if I give a `docsize` of e.g. 20, I want additional_fields to equal the string `,"value0":0,"value1":0,"value2":0,"value3":0,"value4":0`, and then it gets tacked onto the JSON body during the write phase.

Is there a way to do this without Expr and MVEL?

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
jshook: Eric, I'm sure there is a way to do this without falling through to `Expr(...)`. I think you might be interested in `CoinFunc(...)` described here: http://docs.nosqlbench.io/#/docs/bindings/funcref_distributions.md

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
jshook: The collection functions have various flavors of parameter signatures to do exactly this type of thing.
Take a look at the top of the http://docs.nosqlbench.io/#/docs/bindings/funcref_collections.md section, as well as https://github.com/nosqlbench/nosqlbench/blob/main/nb/src/main/resources/examples/bindings-collections.yaml for some ideas. 

If you think of the content as the result of some HOF constructions, it should wire together relatively easily. I'm happy to jump on a call and workshop this too.

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-basic.yaml
```diff
@@ -0,0 +1,226 @@
...
+        Content-Type: "application/json"
+        tags:
+          name: rampup-lt
+
+  - name: rampup-and
+    tags:
+      phase:  rampup-read
+      filter: and
+    statements:
+      - rampup-and: GET <<protocol:http>>://{weighted_hosts}:<<stargate_port:8082>><<path_prefix:>>/v2/namespaces/<<keyspace:docs_search_basic>>/collections/<<table:docs_collection>>?where=%7B%22match1%22%3A%7B%22%24lt%22%3A1%7D%2C%22match2%22%3A%7B%22%24eq%22%3A%22match%22%7D%2C%22match3%22%3A%7B%22%24eq%22%3A%20true%7D%7D&page-size=<<page-size,3>>&fields=<<fields,%5b%5d>>
```
ivansenic, line 155: would be great if you can add as a comment the request lines a decoded query params like (in both files and all statements):

```
# where={"match1":{"$lt":1},"match2":{"$eq":"match"},"match3":{"$eq": true}}&page-size=<<page-size,3>>&fields=<<fields,[]>>
```

ivansenic: And do we need all three here? Would two be fine in fact? Same for OR.. WDYT?

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-basic.yaml
```diff
@@ -0,0 +1,226 @@
...
+        Content-Type: "application/json"
+        tags:
+          name: rampup-and
+
+  - name: rampup-or
+    tags:
+      phase:  rampup-read
+      filter: or
+    statements:
+      - rampup-or: GET <<protocol:http>>://{weighted_hosts}:<<stargate_port:8082>><<path_prefix:>>/v2/namespaces/<<keyspace:docs_search_basic>>/collections/<<table:docs_collection>>?where=%7B%22%24or%22%3A%5B%7B%22match1%22%3A%7B%22%24lt%22%3A1%7D%7D%2C%7B%22match2%22%3A%7B%22%24eq%22%3A%22match%22%7D%7D%2C%7B%22match3%22%3A%7B%22%24eq%22%3Atrue%7D%7D%5D%7D&page-size=<<page-size,3>>&fields=<<fields,%5b%5d>>
```
ivansenic, line 168: Let's add another case, OR where one condition is matching and one not? wdyt? or is this advanced stuff?

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
```diff
@@ -0,0 +1,315 @@
+# nb -v run driver=http yaml=http-docsapi-search-basic tags=phase:schema stargate_host=my_stargate_host auth_token=$AUTH_TOKEN
```
ivansenic, line 1: yaml=http-docsapi-search-basic?

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
```diff
@@ -0,0 +1,315 @@
...
+        Content-Type: "application/json"
+        tags:
+          name: rampup-mem-and
+
+  - name: rampup-mem-or
+    tags:
+      phase:  rampup-read
+      filter: mem-or
+    statements:
+      - rampup-mem-or: GET <<protocol:http>>://{weighted_hosts}:<<stargate_port:8082>><<path_prefix:>>/v2/namespaces/<<keyspace:docs_search_advanced>>/collections/<<table:docs_collection>>?where=%7B%22%24or%22%3A%5B%7B%22match1%22%3A%7B%22%24lt%22%3A1%7D%7D%2C%7B%22match3%22%3A%7B%22%24exists%22%3Atrue%7D%7D%5D%7D&page-size=<<page-size,3>>&fields=<<fields,%5b%5d>>
```
ivansenic, line 179: hmm `"$exists":true` is not in-memory one, only `$exists":false`, you can make up a field: `{"not_existing":{"$exists":false}}`..

Or let again specify, there are three types:
* Cassandra ones (lt, gt)
* In memory single row checked ($in, $exists: true)
* In memory whole document ($ne, $nin, $exists: false)

So I am not sure what case you wanted here..

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
```diff
@@ -0,0 +1,315 @@
+# nb -v run driver=http yaml=http-docsapi-search-basic tags=phase:schema stargate_host=my_stargate_host auth_token=$AUTH_TOKEN
+
+description: |
```
ivansenic, line 3: Can we add an `.MD` for each new scenario, as we did for the CRUD cases?

ivansenic (commented):
Ohhh man this is so awesome.. Can not wait to see the results.. Looking even more forward to that 18th Nov now :smile:

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
EricBorczuk: I was actually taking from the document we have:
```Search with $or (C*-supported + one that does not evaluate on missing)```

Comment (ivansenic) [reply driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml:179]: If this is the case, I am fine!

File: driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml (hunk 5 above)
Comment (EricBorczuk) [reply driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml:1]: My bad, copypasta

File: driver-http/src/main/resources/activities/documents-api/http-docsapi-search-basic.yaml (hunk 3 above)
Comment (EricBorczuk) [reply driver-http/src/main/resources/activities/documents-api/http-docsapi-search-basic.yaml:155]: Yea I think I could simplify this down to two filters

File: driver-http/src/main/resources/activities/documents-api/http-docsapi-search-basic.yaml [hunk 8]
```diff
@@ -0,0 +1,226 @@
...
+
+description: |
+  This workload emulates basic search operations for the Stargate Documents API.
+  During the rampup phase, it generates documents, writes them to a table, and then warms up the search paths.
+  During the main phase it performs various basic search filters and times their execution.
+  Note that stargate_port should reflect the port where the Docs API is exposed (defaults to 8082).
+
+scenarios:
+  schema:         run driver=http tags==phase:schema threads==<<threads:1>> cycles==UNDEF
+  rampup-write:   run driver=http tags==phase:rampup-write cycles===TEMPLATE(docscount,10000000) docsize=TEMPLATE(docsize,15) match-ratio=TEMPLATE(match-ratio,0.01) threads=<<threads:auto>> errors=timer,warn
```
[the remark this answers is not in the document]
EricBorczuk: Yes, that's exactly right - by setting up a ratio of 1%, the same 1 of 100 documents will match the relevant search criteria every time

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-basic.yaml
```diff
@@ -0,0 +1,226 @@
...
+  gender: WeightedStrings('M:10;F:10;O:1')
+  full_name: FullNames()
+  married: ModuloToBoolean()
+  city: Cities()
+  country_code: CountryCodes()
+  lat: Uniform(-180d, 180d)
+  lng: Hash() -> long; Uniform(-180d, 180d)
+  friend_id: Add(-1); ToHashedUUID(); ToString() -> String
+
+  match1: Expr('if (cycle % (int)(1 / <<match-ratio>>) == 0) { return 0; } return (int)(Math.random() * 10000 + 1000);') -> long
```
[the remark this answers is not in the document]
EricBorczuk: There's a way to do it without Expr as well, which I think will be way cleaner :) looking into that today

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-basic.yaml
```diff
@@ -165,6 +167,7 @@ blocks:
       phase:  rampup-read
       filter: or
     statements:
+      # where={"$or":[{"match1":{"$lt":1}},{"match2":{"$eq":"match"}},{"match3":{"$eq":true}}]}
```
ivansenic: let's also decrease here to two conditions only to align with `and` and  `or-single-match`

ivansenic: and same for main of course,

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.md
```diff
@@ -0,0 +1,67 @@
...
+  "match1": 0, // or 1000
+  "match2": "true", // or "false"
+  "match3": true // or false
+}
+```
+
+## Filters tested
+
+The advanced search workload can test the following `where` clauses:
+- in: match1 IN [0]
```
ivansenic, line 53: can we have expression in the `` looks nicer and more readable:

* in: `match1 IN [0]`

(same for the other readme)

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
```diff
@@ -16,7 +16,7 @@ description: |
 # complex3: (match1 LTE 0 AND match2 EQ "true") OR (match2 EQ "false" AND match3 EQ true)
 scenarios:
   schema:             run driver=http tags==phase:schema threads==<<threads:1>> cycles==UNDEF
-  rampup-write:       run driver=http tags==phase:rampup-write cycles===TEMPLATE(docscount,10000000) docsize=TEMPLATE(docsize,15) match-ratio=TEMPLATE(match-ratio,0.01) threads=<<threads:auto>> errors=timer,warn
+  rampup-write:       run driver=http tags==phase:rampup-write cycles===TEMPLATE(docscount,10000000) docpadding=TEMPLATE(docpadding,15) match-ratio=TEMPLATE(match-ratio,0.01) threads=<<threads:auto>> errors=timer,warn
```
ivansenic, line 19: you wrote docs padding is default to zero, but here is still 15? seems wrong?

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
```diff
@@ -50,11 +50,11 @@ bindings:
...
 
-  match1: Expr('if (cycle % (int)(1 / <<match-ratio>>) == 0) { return 0; } return (int)(Math.random() * 10000 + 1000);') -> long
-  match2: Expr('if (cycle % (int)(1 / <<match-ratio>>) == 0) { return "true" } return "false";') -> String
-  match3: Expr('if (cycle % (int)(1 / <<match-ratio>>) == 0) { return "true" } return "false";') -> String
+  match1: Identity(); CoinFunc(<<match-ratio>>, FixedValue(0), FixedValue(1000))
+  match2: Identity(); CoinFunc(<<match-ratio>>, FixedValue("true"), FixedValue("false"))
+  match3: Identity(); CoinFunc(<<match-ratio>>, FixedValue("true"), FixedValue("false"))
   
-  additional_fields: Expr('str="";for (int i=0; i < TEMPLATE(docsize,15) - 15; i++) { if (str == "") { str = "," }; str+="\"value"+i+"\":0"; if (i < TEMPLATE(docsize,15) - 16) {str += ","}} return str;') -> String
+  additional_fields: ListSizedStepped(<<docpadding:0>>,Template("\"{}\":{}",Identity(),Identity())); ToString(); ReplaceAll('\[\"', ',\"'); ReplaceAll('\[', ''); ReplaceAll('\]', '') -> String
```
ivansenic, line 57: have you tested this? can you confirm that docs written are larger?

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-basic.yaml
```diff
@@ -153,8 +153,8 @@ blocks:
       phase:  rampup-read
       filter: and
     statements:
-      # where={"match1":{"$lt":1},"match2":{"$eq":"match"}}
-      - rampup-and: GET <<protocol:http>>://{weighted_hosts}:<<stargate_port:8082>><<path_prefix:>>/v2/namespaces/<<keyspace:docs_search_basic>>/collections/<<table:docs_collection>>?where=%7B%22match1%22%3A%7B%22%24lt%22%3A1%7D%2C%22match2%22%3A%7B%22%24eq%22%3A%22match%22%7D%7D&page-size=<<page-size,3>>&fields=<<fields,%5b%5d>>
+      # where={"match1":{"$lt":1},"match2":{"$eq":"true"}}
```
ivansenic, line 156: hah, adding the comment unveiled a bug :+1:

ivansenic (approved):
I am giving approval, but please fix stuff mentioned in the comments :+1:

driver-http/src/main/resources/activities/documents-api/http-docsapi-search-advanced.yaml
EricBorczuk: Yea I tested this (for the basic test)

jshook (commented):
Can we consider using the `URLENCODE[[...]]` on the URIs in order to make the predicates more readable?

jshook (approved):
That looks good. Lots of nested syntax! But at least we can read the predicate it and infer intent from the yaml.

EricBorczuk:
Seems like a few tests are failing, such as `testCycleRateChangeNewMetrics` - is this a known issue? I don't think it would have to do with my changes at all

jshook (commented):
This looks good. I'll go ahead and merge it.

jshook:
Yes, this is a known issue. It is unrelated to your changes. 
It appears that the environment where github actions run is being clamped down incrementally over time. This makes the integration tests, which verify timely adjustment of rate limiting logic, sensitive. Basically, the test doesn't seem to be able to run at a target rate of 5K ops/s reliably. This is both a host env issue and an algorithm issue. The rate limiter and/or the tests for it need to be made to "scale down" to ultimately solve this.

PR #381, opened 2021-11-11
Pulsar: Workaround an ordering bug in the Pulsar client

lhotari:
- see https://github.com/apache/pulsar/pull/12456 for details about the Pulsar client ordering bug.
  - the bug applies to the use of receiveAsync and the workaround
    is to avoid the use of receiveAsync until the fix is included in the client
- receiveAsync was introduced previously to workaround the bug https://github.com/apache/pulsar/issues/9921
  - that issue isn't critical


jshook (commented):
LGTM

PR #385, opened 2021-11-18
add support for scenarios from an http service

tjake:


engine-cli/src/test/java/io/nosqlbench/engine/cli/NBCLIScenarioParserTest.java
```diff
@@ -125,6 +125,16 @@ public void testThatFullyQualifiedScenarioFilesAreSupported() {
         assertThat(cmds.size()).isGreaterThan(0);
     }
 
+    @Test
+    public void testThatScenarioUrlsAreSupported() {
+        //TODO: This might change?
+        String urlScenario = "https://gist.githubusercontent.com/tjake/0106848762e078f0ef886c1a097b3433/raw/4302fd0ab2d34bd088d973cdd52ed7711782f757/scenario-test.yaml?foo=bar";
+
```
jshook, line 132: It would be better to have this point at the project repo rather than a gist.

jshook (approved):
As long as the other NBIO tests are happy with this, I think it's good.

PR #387, opened 2021-11-25
fix http driver default scenarios

XN137:
in https://github.com/nosqlbench/nosqlbench/pull/306 we seem to have forgotten to update the default scenrios accordingly

(schemas are now created by http POST)

wondering: 
could the build add some validation of these values to prevent similar errors in the future?

could statements have more static checks around driver compatiblity?
for example some statement definitions could require a driver of the "http" family... others from the "cql" family.
As is you get a weird looking runtime error if specifying the wrong driver i think.

### DIFF
--- a/driver-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
+++ b/driver-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
@@ -7,7 +7,7 @@ description: |
 
 scenarios:
   default:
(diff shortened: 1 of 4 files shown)


jshook (commented):
LGTM,

On the request to have more validation around op compatibility, I think we should try. However, there is not yet a mechanism which would allow this to be done as a separate "type checking" action. Consider for example that the stdout activity works with *all* statement forms, and there is no set of rules that can determine between a syntax error and an incompatible choice of driver (yet?)

PR #390, opened 2021-12-08
[Pulsar] Add way to calculate end-to-end operation latency from event time or message property

lhotari:
When testing Pulsar Functions end-to-end latency, it is necessary to be able to calculate the end-to-end latency metrics for the complete end-to-end operation.

This PR adds support for calculating the end-to-end operation latency from either
* message publish time (default when using optype of `ec2-msg-proc-consume`)
* message event time
* message property called "e2e_starting_time" (milliseconds since epoch)

The calculation method should be specified in the `e2e_starting_time_source` document level parameter.
When `e2e_starting_time_source` is set, it configures the end-to-end starting time source. The supported values:
* `message_publish_time` 
* `message_event_time` 
* `message_property_e2e_starting_time`

Pulsar Functions copy the properties from the incoming message by default. This makes `message_property_e2e_starting_time` suitable for calculating end-to-end operation latency for chained Pulsar Functions where one incoming message maps exactly to one output message.

### DIFF
--- a/driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/EndToEndStartingTimeSource.java
+++ b/driver-pulsar/src/main/java/io/nosqlbench/driver/pulsar/ops/EndToEndStartingTimeSource.java
@@ -1,0 +1,8 @@
+package io.nosqlbench.driver.pulsar.ops;
+
+public enum EndToEndStartingTimeSource {
+    NONE, // no end-to-end latency calculation
+    MESSAGE_PUBLISH_TIME, // use message publish timestamp
+    MESSAGE_EVENT_TIME, // use message event timestamp
+    MESSAGE_PROPERTY_E2E_STARTING_TIME // use message property called "e2e_starting_time" as the timestamp
+}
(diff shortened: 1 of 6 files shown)


yabinmeng:
@lhotari LGTM.  Can you please update README? In particular, section 3.2 and chapter 8? Thank you!

jshook (commented):
@lhotari please review the previous changes which added payloadRttFieldFunc to see if these are doing practically the same thing. If so, adding this change would mean we have *three* different ways to calculate some form of end-to-end RTT. I'd prefer we not duplicate functionality on this if it is the same intent.

lhotari:
> @lhotari LGTM. Can you please update README? In particular, section 3.2 and chapter 8? Thank you!

@yabinmeng I've added docs to section 3.2.

lhotari:
> @lhotari please review the previous changes which added payloadRttFieldFunc to see if these are doing practically the same thing. If so, adding this change would mean we have _three_ different ways to calculate some form of end-to-end RTT. I'd prefer we not duplicate functionality on this if it is the same intent.

`payloadRttFieldFunc` is different. It takes the time source for calculating the end-to-end latency from the payload. 

I don't see three different ways for calculating some form of end-to-end RTT. I think there's only two: the one in this PR that uses the `e2e_msg_latency` histogram and the payloadRttFieldFunc feature which added the `payload_rtt` histogram. I agree that it's not optimal that there are multiple ways for calculating end-to-end RTT. Perhaps some further unification could be made in the future. 

This PR improves the existing `e2e_msg_latency` histogram solution so that it is no more tied to using `optype` of `ec2-msg-proc-consume` . This PR introduces a concept of "end-to-end operation starting time source" so that various use cases can be supported while continuing to use the `e2e_msg_latency` histogram.

/cc @mfleming @kijanowski

jshook (approved):
Merging for an early release tomorrow

Issue #388, opened 2021-12-01
Connecting to MongoDB atlas instance using NB driver produces error


EricBorczuk:
**Describe the bug**
When calling NB with a mongodb connection string that represents an atlas instance, an error occurs.

I believe this could be due to use of an outdated version of the mongodb driver, but there could be something more complex going on...

**To Reproduce**
1. run ``java -jar nb.jar nosqlbench/driver-mongodb/src/main/resources/activities/mongodb-search-basic rampup-write connection='mongodb+srv://eric:<password>@serverlessinstance0.ba5rr.mongodb.net' database=testdb docscount=100000 match-ratio=0.1```

2. An error occurs: ```A TXT record is only permitted to contain the keys [authsource, replicaset], but the TXT record for 'serverlessinstance0.ba5rr.mongodb.net' contains the keys [loadbalanced, authsource]```

**What was Expected**
Connection should occur successfully and the scenario should proceed.

**Additional context**
OS: macOS mojave
NB version: 4.15.63

**Screenshots, if applicable**

PR #393, opened 2021-12-13
Update mongo-driver-sync to 4.4.0

EricBorczuk:
Resolves https://github.com/nosqlbench/nosqlbench/issues/388

After bumping version of `mongo-driver-sync` to 4.4.0 (latest), I can run workloads against a mongoDB atlas instance. I've tested that this change works with the Mongo workloads that I've added in the past, and tests pass.


virtdata-lang/src/main/java/io/nosqlbench/virtdata/lang/generated/VirtDataBaseListener.java
```diff
@@ -1,4 +1,4 @@
-// Generated from VirtData.g4 by ANTLR 4.8
+// Generated from VirtData.g4 by ANTLR 4.9.2
```
EricBorczuk, line 1: This all change automatically after I ran `mvn package`

jshook (approved):
LGTM!

PR #403, opened 2022-01-14
Remove extraneous System.out.println statement

pingtimeout:
It was probably committed by accident


jshook (approved):
yes, thanks for fixing this!

PR #411, opened 2022-01-20
Logging: do not set ROOT level to ALL

eolivelli:
Motivation:
During some Pulsar performance tests I have observed some significant hotspots in Pulsar client code (especially in the code that deals with writing and reading to/from the network) due to the fact that "logger.isDebugEnabled()" and "logger.isTraceEnabled()" returns always true. This fact in turn triggers the execution of "logger.debug()" and "logger.trace()", that needs a good amount of method calls and memory allocations that we want to prevent especially in the hottest paths.

The reason is that the ROOT logger in NB is configured with Level.ALL.

Changes:

Configure the ROOT logger the same way as the File level, this is because the fileLevel is supposed to show more than the console.
We already have a getEffectiveFileLevel() that computes the most verbose (specific in log4j2 terms) logger level.

Testing:
I have tested manually this patch, with these results:
1. running NB without flags results in fileLevel/rootLevel: INFO, consoleLevel: WARN
2. running NB with '-v' corresponds to fileLevel/rootLevel: INFO, consoleLevel: INFO
3. running NB with '-vv' corresponds to fileLevel/rootLevel: DEBUG, consoleLevel: DEBUG
4. running NB with '-vvv' corresponds to fileLevel/rootLevel: TRACE, consoleLevel: TRACE

I have verified that internally the code sees the expected answers to isDebugEnabled() and isTraceEnabled() according to the requested level from the CLI.

### DIFF
--- a/engine-core/src/main/java/io/nosqlbench/engine/core/logging/LoggerConfig.java
+++ b/engine-core/src/main/java/io/nosqlbench/engine/core/logging/LoggerConfig.java
@@ -66,7 +66,6 @@ public class LoggerConfig extends ConfigurationFactory {
     private final String logfilePattern = DEFAULT_LOGFILE_PATTERN;
     private NBLogLevel fileLevel = NBLogLevel.DEBUG;
 
-    public static final Level ROOT_LOG_LEVEL = Level.ALL;
     private Map<String, String> logLevelOverrides = new LinkedHashMap<>();
     private Path loggerDir = Path.of("logs");
     private String sessionName;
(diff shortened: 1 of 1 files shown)


engine-core/src/main/java/io/nosqlbench/engine/core/logging/LoggerConfig.java
```diff
@@ -116,9 +115,20 @@ Configuration createConfiguration(final String name, ConfigurationBuilder<BuiltC
...
+
+        // configure the ROOT logger the same way as the File level
+        // this is because the fileLevel is supposed to show more than the console
+
+        // therefore, it is very important that the ROOT level is as much specific as possible
+        // because NB code and especially third party libraries may rely on logging guards (if logger.isDebugEnabled())
+        // to reduce memory allocations and resource waste due to debug/trace logging
+        // if you set ROOT to ALL or to TRACE then you will trigger the execution of trace/debugging code
+        // that will affect performances and impact on the measurements made with NB
+        Level rootLoggingLevel = fileLevel;
```
lhotari, line 130: wouldn't `rootLoggingLevel` have to be either `fileLevel` or `consoleLevel`, picking the finer grain one?

lhotari, line 130:
```suggestion
        Level rootLoggingLevel = fileLevel.intLevel() > consoleLevel.intLevel() ? fileLevel : consoleLevel;
```

lhotari (changes requested):

engine-core/src/main/java/io/nosqlbench/engine/core/logging/LoggerConfig.java
lhotari: Oh sorry, I didn't notice your comment "We already have a getEffectiveFileLevel() that computes the most verbose (specific in log4j2 terms) logger level."

lhotari: This is fine.

lhotari: not needed since "We already have a getEffectiveFileLevel() that computes the most verbose (specific in log4j2 terms) logger level."

lhotari (approved):
LGTM

engine-core/src/main/java/io/nosqlbench/engine/core/logging/LoggerConfig.java
eolivelli: it is already computed...see here:
```
private NBLogLevel getEffectiveFileLevel() {
        if (fileLevel.isGreaterOrEqualTo(consoleLevel)) {
            return fileLevel;
        } else {
            return consoleLevel;
        }
    }
```

jshook (commented):
I think this looks good.

PR #412, opened 2022-01-25
Pulsar: support Schema.AUTO_CONSUME() and support KeyValue for RTT tracking feature

eolivelli:
Modifications:
- support schema.type=AUTO_CONSUME
- unwrap the "value" inside the KeyValue structure for the RTT tracking


eolivelli:
@jshook we need this and possibly a release on the 4.15.x release line (4.15.84?)

could you please take a look ?
I tested this with @mfleming

PR #419, opened 2022-02-03
implement driver-cql-scylla-shaded

dkropachev:
Implement driver for [ScyllaDB](https://www.scylladb.com/), which similar to `driver-cqld3-shaded`, but have some syntax difference and using [scylla java driver](https://mvnrepository.com/artifact/com.scylladb/scylla-driver-core)


jshook (commented):
@dkropachev Thanks for the PR. We need to talk about the upcoming nosqlbench changes in order to do it right. I'll start an issue for discussion. (We're working on removing shaded and older driver versions entirely from the project once the j17-preview branch is ready, so some adjustments need to be made.)

jshook:
@dkropachev I've added #421 as a way to explain and discuss what needs to be done for an effective PR. Here are the current options that I can think of:
1) merge this after review, with separate work to occur on the j17-preview branch.
2) Prepare a driver on the j17-preview branch.
3) Use the CQLd4 driver on the j17-preview branch.

In any case, we won't be using shading on the main branch once the j17-preview branch is ready for merge. Perhaps we need to talk about the specific differences you have in client behavior and why it has to be a different _re-instanced_ copy of a whole driver tree.

dkropachev:
> @dkropachev I've added #421 as a way to explain and discuss what needs to be done for an effective PR. Here are the current options that I can think of:
> 
> 1. merge this after review, with separate work to occur on the j17-preview branch.
> 2. Prepare a driver on the j17-preview branch.
> 3. Use the CQLd4 driver on the j17-preview branch.
> 
> In any case, we won't be using shading on the main branch once the j17-preview branch is ready for merge. Perhaps we need to talk about the specific differences you have in client behavior and why it has to be a different _re-instanced_ copy of a whole driver tree.

Regarding differences, in major part `Datastax` and `ScyllaDB` drivers are similar, but there are crucial differences:
1. ScyllaDB supports shard awareness, meaning that driver holds separate connection to each shard and also have special logic that allows it to know to which exactly shard query should go, which allows it to send queries directly to the shard, skipping coordinator. This feature optimize network usage inside the cluster and make it more productive.
2. There are some features that exist in Datastax and do not exist in ScyllaDB, which is could be disabled via ANTLR syntax schema. One example is stored FUNCTIONS
3. There are some features that do not exist in Datastax and exist in ScyllaDB, such as a. Caching control ([BYPASS CACHE](https://docs.scylladb.com/getting-started/dml/#bypass-cache), and disable cache per table) and b. [ICS](https://www.scylladb.com/2021/04/28/incremental-compaction-2-0-a-revolutionary-space-and-write-optimized-compaction-strategy/)

Because of these differences I think that It would be better to keep both drivers separately.
Regarding `new driver API`, I do not want to put an extra work on you, so, If `new driver API`/`j17-preview` is far from been ready, that it would be great to have this PR merged and later prepare driver using new API on `j17-preview`.
At worst you will merge `j17-preview` to `main` without `ScyllaDB` driver.

Regarding testing, I will take a look at what you have there and do similar things on this driver, also If you have `testing protocol` beside code, let me know, I will follow it to make sure it is fully functional.

What is the status of `j17-preview`, is `cqld4` functional there ? Can I start looking at it ?

jshook:
@dkropachev Yes, the cqld4 driver is functional. The key APIs to look at are DriverAdapter and its implementations. You'll find a different set of supporting types that I would have called "missing tools." If you study the Cqld4DriverAdapter and the DynamoDBDriverAdapter, you'll see how they are used. Specifically, these types are all more targeted and prescriptive around how to construct ops: DriverAdapter, OpMapper, OpDispenser, and ParsedOp. Also, native driver state (the object graph) associated with an instance of a driver must now live in a DriverSpaceCache (which can be initialized as needed via a DriverAdapter delegate method).

If you study the two initial implementations above, it should start to make sense. Ideally these are all in a developer docs site soon with some supporting examples. 

If you are ok with using the j17-preview branch, I think it will save us both effort, since the API changes are causing some structural divergence from main.

dkropachev:
> @dkropachev Yes, the cqld4 driver is functional. The key APIs to look at are DriverAdapter and its implementations. You'll find a different set of supporting types that I would have called "missing tools." If you study the Cqld4DriverAdapter and the DynamoDBDriverAdapter, you'll see how they are used. Specifically, these types are all more targeted and prescriptive around how to construct ops: DriverAdapter, OpMapper, OpDispenser, and ParsedOp. Also, native driver state (the object graph) associated with an instance of a driver must now live in a DriverSpaceCache (which can be initialized as needed via a DriverAdapter delegate method).
> 
> If you study the two initial implementations above, it should start to make sense. Ideally these are all in a developer docs site soon with some supporting examples.
> 
> If you are ok with using the j17-preview branch, I think it will save us both effort, since the API changes are causing some structural divergence from main.

Great, what do you think of the merging this PR into `master` and creating an issue and separate PR to implement scylla driver support on `j17-preview` ?

jshook:
I'm ok with that, however the goal is to merge j17-preview into main very soon. So that shifts the focus to the j17-preview implementation, since you would lose access to the driver without it. However, getting this into a current release would allow it to be used even if you version locked to a specific release. So I suppose both are valid.

dkropachev:
> I'm ok with that, however the goal is to merge j17-preview into main very soon. So that shifts the focus to the j17-preview implementation, since you would lose access to the driver without it. However, getting this into a current release would allow it to be used even if you version locked to a specific release. So I suppose both are valid.

@jshook , could you please take a look at - https://github.com/nosqlbench/nosqlbench/pull/427

jshook:
Closing this in lieu of the approach described in #427 . Shaded artifacts will not be included by default in the main NoSQLBench artifacts, although externally hosted drivers can be loaded at runtime using the steps described there.

PR #498, opened 2022-03-08
Update aplv2 jdbc

jshook:
@szimmer1 Do you approve adding the APLv2 license to these files and assigning copyright to the NoSQLBench project?
I'm cleaning up missing license files.


szimmer1:
Yes, go ahead thanks for adding

PR #565, opened 2022-04-05
Pulsar: fix race in creating Pulsar Consumer and upgrade to Pulsar 2.8.3

eolivelli:
**Motivation**

If you start the Pulsar Driver, PulsarConsumerOp, with multiple threads you end up in creating multiple Consumers but then you are going to use only one of them.
This is a problem in case of Pulsar SHARED or KEY_SHARED subscriptions, because the zombie Consumer will pre-fetch messages that won't ever be processed, an so NB will never be able to fully consume a topic

**Changes**

- Use "computeIfAbsent"
- Upgrade Pulsar Client to 2.8.3 (not strictly needed)

### DIFF
--- a/driver-pulsar/pom.xml
+++ b/driver-pulsar/pom.xml
@@ -18,7 +18,7 @@
     </description>
 
     <properties>
-        <pulsar.version>2.8.2</pulsar.version>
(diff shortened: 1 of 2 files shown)


lhotari (approved):
Good catch!

PR #618, opened 2022-06-30
Fix indentation in baselines2/cql-keyvalue2.yaml for schema-astra block

hemidactylus:
Workload file cql-keyvalue2.yaml had an indentation too much for the block "schema-astra". That made the "block" not be treated like a block, so for instance a command such as `nb5 cql-keyvalue2 astra [...]` would fail immediately with error
```
javax.script.ScriptException: io.nosqlbench.nb.api.errors.BasicError: There were no active statements with tag filter 'block:schema-astra', since all 7 were filtered out.
```

Reducing the indentation of the block bringing it on par with the other blocks fixed the issue.


jshook:
Thanks!

PR #698, opened 2022-08-12
Set base to 17-jdk so this image can run on arm64

weideng1:
This allows nosqlbench docker image to run natively on Apple M1 and other ARM Linux environment


weideng1:
I noticed that the new base image is 30MB larger (227MB vs. 197MB previously) for linux/amd64, but I think it's a worthwhile change so that NB as a performance tool can run most efficiently in ARM64 environments without relying on a CPU emulator (which will slow things down significantly).

PR #707, opened 2022-08-24
some soft links

phact:


jshook (approved):
I am approving this provisionally to work around current issues in the caller, BUT we need to wind this back once we fix the calling system.

Issue #561, opened 2022-04-04
Warn user if they use "count" for the v4.17+ error handler parameters


hemidactylus:
The parameter `errors=count` has become `errors=counter` starting with v4.17 onward.

While the translation happens automatically, it would be good to issue a warning (perhaps linking to the docs page) to help users transition to the new error handler usage.

(possibly related to other automatic conversion of error-handler parameters, if any occur).

jshook:
merged in main

PR #737, opened 2022-10-07
Notify user when using error=count vs. counter.

jeffbanks:
Logic to support #561 

Addresses warning as requested for use of error handler `count` vs. `counter`.

Includes unit test coverage w/ addition of use of `NBMock` for mocking a LogAppender.


jshook (approved):
LGTM, Thanks for taking the initiative on test coverage.

PR #779, opened 2022-11-14
Shutdown hooks

jshook:
This is to unblock an important resource leak fix for driver adapters. Specifically it uses the internal Shutdownable decorator interface to enable auto-shutdown of resources. This will unblock a change needed for the Pulsar adapter.


adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/Cqld4Space.java
```diff
@@ -299,4 +300,8 @@ public static NBConfigModel getConfigModel() {
 
     }
 
+    @Override
+    public void shutdown() {
+        this.getSession().close();
```
jeffbanks, line 305: Should we consider logging an error or warn level message here?  I believe this implements auto-closeable but throws an exception.

nbr-examples/src/test/resources/scripts/examples/extension_histostatslogger.js
```diff
@@ -18,16 +18,14 @@ activitydef = {
     "alias" : "testhistostatslogger",
     "driver" : "diag",
     "cycles" : "50000",
-    "threads" : "20",
-    "interval" : "2000",
-    "targetrate" : "10000.0",
+    "threads" : "5",
```
jeffbanks, line 21: Looks like this one snuck in this PR, but makes sense as we wanted this anyway.

adapters-api/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/DriverSpaceCache.java
```diff
@@ -64,4 +66,8 @@ public S get(String name) {
         return cache.computeIfAbsent(name, newSpaceFunction);
     }
 
+    public Map<String,S> getElements() {
+        return Collections.unmodifiableMap(cache);
```
jeffbanks, line 70: I like the unmodifiable usage.

jeffbanks (approved):
LGTM overall, just one comment for consideration.

adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/Cqld4Space.java
jshook: As in, wrap any exceptions from `.close()` and throw an error there? We're not using try-with-resources here, so I don't know if auto-closeable can help much.

jshook: But I do agree that the code is not defensive enough without an exception handler.

jeffbanks: Right, I was referring to the usage of the session instance itself and if `null` don't attempt to close, while wrapping the close() w/ try/catch to ensure we log out that it couldn't shutdown fully for some reason.

The session implements `AsyncAutoCloseable` so I was thinking about how the lifecyle may have been somewhat controlled prior to the `shutdown` call on this Space instance, but we never really know how that is used.  😄

jshook: I'm reworking this code to re-use auto-closeable at the space level. Also adding integrated tests to ensure that errors are propagated correctly. After some reflection, I think we should not do anything special for errors here as we already have a configurable error handler for such.

jshook: A nasty side-effect of using auto-closable is that developers are warned about not using try-with-resources. Switching this back to Shutdownable.

jshook: ... except that this ^ only affects the core code, and so not a problem for devs. The latest changes use AutoCloseable and also keep the adapter generally stateless except for the spaces, so close() is not called through the adapter, but directly on the spaces visible to the activity. Added tests to provide coverage as well.

phact (approved):

PR #781, opened 2022-11-15
Sporadic test failures w/ gh actions

jeffbanks:
The source of at least one issue has been identified as a race condition between a single-threaded activity with an exception, and the owning type which manages the executor for it. In effect, the exception on the thread causes it to stop, which allows the active wait on executor service shutdown to continue. The following logic in that code path looks at a signal variable to see if the service was stopped exceptionally or normally. When the thread is stopped, the exception handler is then in a race to update the error variable via callback before the activity shutdown logic completes.

The impact of this will be low in practice, only possible when there is a single-threaded exception. However, the result will manifest as a false-positive in the affirmative, meaning that some tests may appear to work when in fact an exception occurred and the scenario logic was not aware of it and signaled zero exit status.

The stability fix for now is to increase the thread count to 2 for the integration tests, as this keeps the thread pool busy long enough for explicit shutdown logic happen-after the callback mechanism. A better fix needs to be implemented, and I'll add details for this in a separate ticket.


jshook (approved):
pushing this through now since log files are being generated

engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLI.java
```diff
@@ -467,13 +468,13 @@ public Integer applyDirect(String[] args) {
...
         executor.execute(scenario);
 
-        while (true) {
-            Optional<ScenarioResult> pendingResult = executor.getPendingResult(scenario.getScenarioName());
-            if (pendingResult.isPresent()) {
-                break;
-            }
-            LockSupport.parkNanos(100000000L);
-        }
+//        while (true) {
```
jeffbanks, line 471: Did you want this to remain in place?

engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/ScenarioResult.java
```diff
@@ -170,8 +156,7 @@ public void reportCountsTo(PrintStream printStream) {
                 }
             } else if (v instanceof Gauge) {
                 Object value = ((Gauge) v).getValue();
-                if (value != null && value instanceof Number) {
-                    Number n = (Number) value;
+                if (value != null && value instanceof Number n) {
```
jeffbanks, line 159: You can remove the null check on this one.

nbr/src/test/java/io/nosqlbench/cli/testing/ExitStatusIntegrationTests.java
```diff
@@ -78,7 +78,7 @@ void testExitStatusOnActivityOpException() {
         invoker.setLogDir("logs/test");
         ProcessResult result = invoker.run("exitstatus_asyncstoprequest", 30,
                 "java", "-jar", JARNAME, "--logs-dir", "logs/test/asyncstop", "--logs-level", "debug", "run",
-                "driver=diag", "cyclerate=10", "op=erroroncycle:erroroncycle=10", "cycles=500", "-vvv"
+                "driver=diag", "threads=2", "cyclerate=10", "op=erroroncycle:erroroncycle=10", "cycles=500", "-vvv"
```
jeffbanks, line 81: Not sure if still relevant but we had this one: `testExitStatusOnActivityThreadException` that was in the older version that would potentially need this threads=2 inclusion.  You may have decided this test was sufficient.  I think the only different was that one threw an unexpected exception instead of using a stop request.  

An `exitstatus_threadexception` I believe.

jeffbanks (commented):
Looks good overall!  I added a few suggestions for consideration.

engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLI.java
jshook: No, this was hackish.

engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/ScenarioResult.java
jshook: done and pushed

nbr/src/test/java/io/nosqlbench/cli/testing/ExitStatusIntegrationTests.java
jshook: I'll keep an eye on it and apply a work-around if needed. My first inclination is to solve the race condition.

PR #829, opened 2022-12-07
Add timeseries test for Docs API

kijanowski:
Goal is to have the same workload (timeseries) across all Stargate APIs (rest, docs, graphql) to make comparisons more direct


jshook (approved):
looks good to me

PR #844, opened 2022-12-12
nosqlbench-843-cqlgen-mv

jshook:


adapter-cqld4/src/main/java/io/nosqlbench/cqlgen/model/CqlModelBuilder.java
```diff
@@ -175,7 +175,11 @@ public void exitOrderDirection(CqlParser.OrderDirectionContext ctx) {
 
     @Override
     public void exitTableOptionItem(CqlParser.TableOptionItemContext ctx) {
-        table.setCompactStorage(ctx.kwCompactStorage() != null);
+        if (table!=null) {
```
jeffbanks, line 178: This looks good, just on line:34, while looking at this `table` reference there was a builder url misspelling, it may be a known thing, but part of a reference so thought I would bring it up.

jeffbanks (approved):
LGTM overall for preventing table ref. NPEs when MVs are involved.

PR #847, opened 2022-12-13
Add NB5 Kafka driver

yabinmeng:
1) Producer with sync and async message send ack.

2) Consumer with 
- auto.commit
- manual commit with specified message receiving batch
- sync and async message recv ack/commit
- 
3) TBD - transaction support is mostly in place but still has some issues.


jshook (approved):
Looks pretty good so far. Merging for further UAT.

eolivelli (commented):
nice work!

PR #854, opened 2022-12-14
Fix Kafka producer transaction issue and add support for subscribing from multiple topics

yabinmeng:
NB yaml files and README are also updated


jshook (approved):
LGTM

PR #863, opened 2022-12-19
nb5 - Issue 446 - Modified the export-docs command and 'build' workflow

MikeYaacoubStax:
The `nb5 export-docs` command has been modernized to generate a zip file in a user specified directory (using --zipfile).
The zip file generated currently includes 2 sub-directories, bindings and drivers, these two sub-directories contain zola parsable markdowns representing the documentation on specific subjects.

The build workflow is triggered on any push to the main branch (A non-release condition needs to be added to the trigger).
The 'build' job in the build workflow exports the zip file and uploads it as an artifact to be downloaded by the 'docs' job.
The docs job downloads and unzips the contents of the documentation zip file, then, using nb-bot, synchronizes the newly exported documentation with the one found in the nosqlbench-build-docs repo, and pushes the changes to that repo.

The nosqlbench-build-docs repo is configured with a workflow of its own to to publish the documentation that has just been pushed to the preview website.

To make sure that the generated markdowns are compatible with zola, we run the raw markdown string through a MutableMarkdown parser that adds the necessary front matter if it is missing.


.github/workflows/build.yml
```diff
@@ -2,39 +2,109 @@ name: build
...
+        with:
+          name: exported-docs
+
+      - name: unzip docs
+        run: unzip exported_docs.zip
+
+      - run: ls -la
+
+      #      - name: download docs
+      #        uses: actions/download-artifact@v1
```
jshook, line 75: We can go ahead and clean up old comments here and further down.

nb-api/src/main/java/io/nosqlbench/api/markdown/aggregator/MutableMarkdown.java
```diff
@@ -69,13 +74,19 @@ private void parseStructure(String rawMarkdown) {
...
                 node=node.getNext();
             }
         }
         if (frontMatter.getTitle()==null || frontMatter.getTitle().isEmpty()) {
-            throw new RuntimeException("The markdown file at '" + this.path.toString() + "' has no heading to use as a title.");
+            if(this.path != null)
+                throw new RuntimeException("The markdown file at '" + this.path.toString() + "' has no heading to use as a title.");
+            else
+                throw new RuntimeException("The markdown string provided has no heading to use as a title.");
         }
```
jshook, line 90: Ok for now, but let's try to add _some_ context for users in these error messages. They shouldn't have to go fishing for where the issue might be. If there are other outputs in the display context to get them focused on the right files, that's good, but it's always important to put essential troubleshooting context into errors.

.github/workflows/build.yml
MikeYaacoubStax: Agreed!

virtdata-userlibs/src/main/java/io/nosqlbench/virtdata/userlibs/apps/VirtDataMainApp.java
```diff
@@ -46,7 +48,7 @@ public static void main(String[] args) {
     @Override
     public int applyAsInt(String[] args) {
         if (args.length == 0) {
-            System.out.println("Usage: app (" + APP_TESTMAPPER + "|" + APP_GENDOCS + "|" + APP_DIAGNOSE +")");
+            System.out.println("Usage: app (" + APP_TESTMAPPER + "|" + APP_BUILDDOCS + "|" + APP_GENDOCS +"|"+ APP_DIAGNOSE +")");
```
jshook, line 51: We can remove the out-moded calls. Basically, *gendocs* should go away.

jshook (approved):
Some minor comments, potentially for additional improvements. This looks good to merge for now, so we can get the ball rolling on the doc updates.

virtdata-userlibs/src/main/java/io/nosqlbench/virtdata/userlibs/apps/VirtDataMainApp.java
MikeYaacoubStax: Done, committing and pushing

nb-api/src/main/java/io/nosqlbench/api/markdown/aggregator/MutableMarkdown.java
MikeYaacoubStax: We could pass a markdown name to the constructor in case we don't have a path to an actual file on the filesystem?

jshook: Or something like that. Let's try to keep it as simple as possible, but you have the right idea.

PR #864, opened 2022-12-19
nosqlbench-861-dryrun

jshook:


jeffreyscarpenter (approved):

adapters-api/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/DryRunOp.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class DryRunOp implements RunnableOp {
+
+    private final Op op;
+
+    public DryRunOp(Op op) {
+        this.op = op;
+    }
+    @Override
+    public void run() {
```
jeffbanks, line 14: Do we want to log anything at debug level, or would that be a bit verbose for these ops?

PR #870, opened 2022-12-20
update stargate workloads for v2 and nb5

jeffreyscarpenter:
Stargate workloads were not working on nb5. Also made updates to fit Stargate v2 architecture (i.e. `restapi_host` instead of `stargate_host`), and to create all schema via APIs instead of CQL.


jshook (approved):
All looks good to me.

PR #878, opened 2022-12-21
nosqlbench-875-categories

jshook:


jeffbanks (approved):
These look good based on our offline conversation about categories.

PR #883, opened 2022-12-22
Nosqlbench 881 fixtest

### LINKED ISSUE #881: stabilize flaky test: ScriptExampleTests.testExtensionHistogramLogger
```
shutdown hook running
[INFO] 
[INFO] Results:
[INFO] 
Error:  Failures: 
Error:    ScriptExampleTests.testExtensionHistogramLogger:185 
Expecting actual:
  1
to be greater than or equal to:
  2
```

jshook:
This closes #881 and makes test logging generally better overall.

### DIFF
--- a/adapter-cqld4/pom.xml
+++ b/adapter-cqld4/pom.xml
@@ -100,7 +100,7 @@
             <plugin>
                 <groupId>org.antlr</groupId>
                 <artifactId>antlr4-maven-plugin</artifactId>
-                <version>4.10.1</version>
+                <version>4.11.1</version>
                 <configuration>
                     <sourceDirectory>src/main/java/io/nosqlbench/cqlgen/grammars
                     </sourceDirectory>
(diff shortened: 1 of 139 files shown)


jeffbanks:
@jshook @MikeYaacoubStax what do you think about us also considering to use parameterized logging for our formatted messages in the future?  Not just test logic, but other areas of the logic as well in nb.

```java
logger.debug("Name: {} Address: {} ", name, address);
```

There are some readability and efficiency advantages to using approach.  Not for this PR specifically, but for our standards moving forward and we can refactor as we go.  wdyt?

MikeYaacoubStax:
> @jshook @MikeYaacoubStax what do you think about us also considering to use parameterized logging for our formatted messages in the future? Not just test logic, but other areas of the logic as well in nb.
> 
> ```java
> logger.debug("Name: {} Address: {} ", name, address);
> ```
> 
> There are some readability and efficiency advantages to using approach. Not for this PR specifically, but for our standards moving forward and we can refactor as we go. wdyt?



> @jshook @MikeYaacoubStax what do you think about us also considering to use parameterized logging for our formatted messages in the future? Not just test logic, but other areas of the logic as well in nb.
> 
> ```java
> logger.debug("Name: {} Address: {} ", name, address);
> ```
> 
> There are some readability and efficiency advantages to using approach. Not for this PR specifically, but for our standards moving forward and we can refactor as we go. wdyt?

@jeffbanks do you mean as opposed to static log messages? If so, I don't see why not, the logger is initialized with the class it's being used in so we could always check that class for those variables.

jeffbanks:
@MikeYaacoubStax 

Really the suggestion is more for use of the parameters as opposed to the message contents being aggregated via string '+' concatenation.  

The logging API will evaluate the statement and only create the error message contents when determined to be used or not.  For example, if set to log level `info` the `debug` log messages won't be constructed when using parameters.

[Here ](https://stackoverflow.com/questions/30975026/log4j-implicit-string-formatting)is a good explanation I found about this topic.

jeffbanks (approved):
LGTM

PR #885, opened 2022-12-23
Added support for a second inline binding format

jshook:
This provides an alternative form for when the templating language conflicts with other tools, like mustache or zola.


jeffbanks (approved):
lgtm

adapter-http/src/main/resources/http.md
```diff
@@ -203,7 +203,7 @@ an op template like `space: {space}`, OR use an inline op field in your op
 template like `space: {(ThreadNumToInteger())}`.
 
 You can use any binding function you want for the space op field. However,
-if you were to assign it something like "space: {{Identity()}}" you would
+if you were to assign it something like "space: {(Identity()}" you would
```
msmygit, line 206: `#trivial` I see there is an open paranthesis "`(`" but, don't see the equivalent closing paranthesis "`)`" and just wanted to call your attention here to validate if this is indeed the right format.

jshook: Thanks, I'll fix this.

jshook: fixed in current code to be pushed.

PR #893, opened 2023-01-04
nosqlbench-440 convert tcp driver from ActivityType to DriverAdapter #440 #441

### LINKED ISSUE #440: convert tcpclint driver from ActivityType to DriverAdapter

### LINKED ISSUE #441: convert tcpserver driver from ActivityType to DriverAdapter e:1d
Implement the tcpserver adapter using the basic logic from the old tcpserver activity type.

MikeYaacoubStax:
Converted TCPServer and TCPClient from ActivityType drivers to DriverAdapter drivers. Related NB Issues are: #440 and #441 .

Both of these drivers can be configured through the driver parameters indicated in their respective markdowns.

The configuration happens through activity parameters ( nb5 run driver=tcpclient .... [DRIVER PARAM1] = x [DRIVER PARAM2] = y.

Unit testing through RawSocketInjector still needs to be added.

Closes #440 
Closes #441


adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpclient/TcpClientAdapterSpace.java
```diff
@@ -0,0 +1,113 @@
...
+import javax.net.SocketFactory;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.Writer;
+import java.net.Socket;
+
+public class TcpClientAdapterSpace {
+
+    private final static Logger logger = LogManager.getLogger(TcpServerAdapterSpace.class);
```
jeffbanks, line 37: Update with class TcpClientAdapterSpace

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpclient/TcpClientAdapterSpace.java
```diff
@@ -0,0 +1,113 @@
...
+    private final NBConfiguration config;
+    Writer writer;
+    private PrintWriter console;
```
jeffbanks, line 40: can be removed, not used.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpclient/TcpClientAdapterSpace.java
```diff
@@ -0,0 +1,113 @@
...
+            .add(
+                Param.optional("format")
+                    .setRegex("csv|readout|json|inlinejson|assignments|diag")
+                    .setDescription("Which format to use.\n" +
+                        "If provided, the format will override any statement formats provided by the YAML. " +
+                        "If 'diag' is used, a diagnostic readout will be provided for binding constructions.")
+            )
+            .add(
+                Param.defaultTo("bindings","doc")
+                    .setDescription("This is a simple way to specify a filter for the names of bindings that you want to use.\n" +
```
jeffbanks, line 104: One option here is to utilize the Java text block.  You can remove concatenation, quotes, etc. for the string and replace with the following:

"""
<body of text>
"""

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpclient/TcpClientDriverAdapter.java
```diff
@@ -0,0 +1,66 @@
...
+import io.nosqlbench.engine.api.activityimpl.uniform.DriverAdapter;
+import io.nosqlbench.engine.api.activityimpl.uniform.DriverSpaceCache;
+import io.nosqlbench.engine.api.activityimpl.uniform.decorators.SyntheticOpTemplateProvider;
+import io.nosqlbench.nb.annotations.Service;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.*;
+import java.util.function.Function;
+import java.util.regex.Pattern;
```
jeffbanks, line 36: unused import

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import javax.net.ServerSocketFactory;
+import javax.net.ssl.SSLServerSocketFactory;
+
+public class TcpServerAdapterSpace implements AutoCloseable{
+    @Override
+    public void close() throws Exception {
```
jeffbanks, line 47: This method should be below the variable declarations.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    private final static Logger logger = LogManager.getLogger(TcpServerAdapterSpace.class);
+    private final NBConfiguration config;
+    Writer writer;
+    private LinkedBlockingQueue<String> queue;
+    private ServerSocket listenerSocket;
+    private final List<Shutdown> managedShutdown = new ArrayList<>();
+    private int capacity=10;
+    private ServerSocketFactory socketFactory;
```
jeffbanks, line 69: This can be removed the instance variable level and used in the method where it used.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+            .add(
+                Param.optional("format")
+                    .setRegex("csv|readout|json|inlinejson|assignments|diag")
+                    .setDescription("Which format to use.\n" +
+                        "If provided, the format will override any statement formats provided by the YAML. " +
+                        "If 'diag' is used, a diagnostic readout will be provided for binding constructions.")
+            )
+            .add(
+                Param.defaultTo("bindings","doc")
+                    .setDescription("This is a simple way to specify a filter for the names of bindings that you want to use.\n" +
```
jeffbanks, line 160: Java text block usage can be used here.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+
+            )
+            .asReadOnly();
+    }
+    private interface Shutdown {
+        void shutdown();
+    }
+    public static class SocketWriter implements Runnable, Shutdown {
+        private final BlockingQueue<String> sourceQueue;
+        private final OutputStream outputStream;
```
jeffbanks, line 173: This one can be moved down to where it used as a local variable.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+        private final OutputStreamWriter writer;
+        private boolean running = true;
+
+
+        public SocketWriter(BlockingQueue<String> sourceQueue, Socket connectedSocket) {
+            this.sourceQueue = sourceQueue;
+            try {
+                outputStream = connectedSocket.getOutputStream();
+                this.writer = new OutputStreamWriter(outputStream);
+                //connectedSocket.shutdownInput();
```
jeffbanks, line 183: Can be removed.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+            }
+        }
+
+        public void shutdown() {
+            this.running = false;
+        }
+
+        @Override
+        public void run() {
+            try (Writer writer = this.writer) {
```
jeffbanks, line 195: Maybe change the name of this `writer` variable to be something else like runWriter or threadWriter to avoid confusion with the instance variable declared above.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+                    try {
+                        Thread.sleep(10);
+                    } catch (InterruptedException ignored) {
+                    }
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+            finally {
+                //TODO: Add the shutdown logic to do after closing the socket
```
jeffbanks, line 215: We should try to do this as part of the PR, it is important to get that closed after creation regardless of exception or not.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+            }
+        }
+
+        @Override
+        public synchronized void flush() throws IOException {
+        }
+
+        @Override
+        public synchronized void close() throws IOException {
+            flush();
```
jeffbanks, line 248: Let's put a `super.flush()` so the base class logic get invoked.
This can thrown an exception so it might be good to catch this and log as an error, and then set the queue to null.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+            this.serverSocket = serverSocket;
+        }
+
+        public void shutdown() {
+            this.running = false;
+        }
+
+        @Override
+        public void run() {
+            try (ServerSocket serverSocket = this.serverSocket) {
```
jeffbanks, line 270: Let's change the name to runServerSocket or threadServerSocket to avoid confusion with the instance variable `serverSocket` defined as an instance variable.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+                while (running) {
+                    serverSocket.setSoTimeout(1000);
+                    serverSocket.setReuseAddress(true);
+                    try {
+                        Socket connectedSocket = serverSocket.accept();
+                        SocketWriter writer = new SocketWriter(queue, connectedSocket);
```
jeffbanks, line 276: Same name change suggestion as we have a writer variable at the instance scope.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+                        //TCPServerActivity.this.managedShutdown.add(writer);
```
jeffbanks, line 277: Dead code can be removed.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,291 @@
...
+                        Thread writerThread = new Thread(writer);
+                        writerThread.setName("SocketWriter/" + connectedSocket);
+                        writerThread.setDaemon(true);
+                        writerThread.start();
+                        logger.info("Started writer thread for " + connectedSocket);
+                    } catch (SocketTimeoutException ignored) {
```
jeffbanks, line 283: I'm wondering for the cases in NB where we ignore exceptions if we should at least put in a logger.debug(...) entry to indicate that it happened in case we need to turn on debug level to diagnose a strange issue.  wdyt?

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerOpDispenser.java
```diff
@@ -0,0 +1,46 @@
...
+public class TcpServerOpDispenser extends BaseOpDispenser<TcpServerOp,TcpServerAdapterSpace> {
+
+    private final LongFunction<TcpServerAdapterSpace> ctxfunc;
+    private final LongFunction<String> outFunction;
+
+    public TcpServerOpDispenser(TcpServerDriverAdapter adapter, ParsedOp cmd, LongFunction<TcpServerAdapterSpace> ctxfunc) {
+        super(adapter,cmd);
+        this.ctxfunc = ctxfunc;
+        LongFunction<Object> objectFunction = cmd.getAsRequiredFunction("stmt", Object.class);
+        LongFunction<String> stringfunc = l -> objectFunction.apply(l).toString();
```
jeffbanks, line 35: nit: this can be `stringFunction` to match `objectFunction`

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerOpDispenser.java
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.nosqlbench.adapter.tcpserver;
+
+import io.nosqlbench.engine.api.activityimpl.BaseOpDispenser;
+import io.nosqlbench.engine.api.activityimpl.uniform.DriverAdapter;
```
jeffbanks, line 20: Can remove DriverAdapater and Op as imports not used.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerOpMapper.java
```diff
@@ -0,0 +1,45 @@
...
+
+
+    public TcpServerOpMapper(TcpServerDriverAdapter adapter, DriverSpaceCache<? extends TcpServerAdapterSpace> ctxcache) {
+        this.ctxcache = ctxcache;
+        this.adapter = adapter;
+    }
+
+    @Override
+    public OpDispenser<TcpServerOp> apply(ParsedOp op) {
+        LongFunction<String> spacefunc = op.getAsFunctionOr("space", "default");
```
jeffbanks, line 39: Nits: spaceFunction and ctxFunction for names.

adapter-tcp/src/main/resources/tcpclient.md
```diff
@@ -41,11 +35,11 @@ Run a stdout activity named 'stdout-test', with definitions from activities/stdo
...
   [Additional parameters may need to be provided](../../../../driver-cql/src/main/resources/ssl.md).
 
-- **host** - this is the name to connect to (remote server IP address)
+- **host** - this is the name to bind to (local interface address)
   - default: localhost
   - dynamic: false
-- **port** - this is the name of the port to connect to (remote server port)
-  - default: 12345
+- **port** - this is the name of the port to listen on
+  - default: 8080
```
jeffbanks, line 42: We may want to verify this port with @jshook , it is a common one across many apps and tooling.

adapter-tcp/src/main/resources/tcpserver.md
```diff
@@ -49,7 +51,7 @@ Run a stdout activity named 'stdout-test', with definitions from activities/stdo
   - default: localhost
   - dynamic: false
 - **port** - this is the name of the port to listen on
-  - default: 12345
+  - default: 8080
```
jeffbanks, line 54: Same as comment above.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
MikeYaacoubStax: This is a good idea, but then when I tried to implement it I discovered that flush() is only an abstract in the base class (java Writer). so it doesn't really do any extra logic in the base class, this also means that our call to "flush()" in close isn't really doing anything right now.

adapter-tcp/src/main/resources/tcpclient.md
MikeYaacoubStax: reverting everything to 12345, if it's the default port for other drivers as well, it should be the same here.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
MikeYaacoubStax: I think extra help is always appreciated when debugging, I added a logger debug message indicating the socket where the exception is thrown.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpclient/TcpClientOpDispenser.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final LongFunction<TcpClientAdapterSpace> ctxFunction;
+    private final LongFunction<String> outFunction;
+
+    public TcpClientOpDispenser(TcpClientDriverAdapter adapter, ParsedOp cmd, LongFunction<TcpClientAdapterSpace> ctxfunc) {
+        super(adapter,cmd);
+        this.ctxFunction = ctxfunc;
+        LongFunction<Object> objectFunction = cmd.getAsRequiredFunction("stmt", Object.class);
+        LongFunction<String> stringFunction = l -> objectFunction.apply(l).toString();
+        cmd.enhanceFuncOptionally(stringFunction,"suffix",String.class,(a, b) -> a+b);
```
jshook, line 34: This is special to stdout, whereas we now have integrated string templating in the form of "prefix{binding}suffix" format support. We should probably remove this op template field to avoid duplicity or confusion.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerAdapterSpace.java
```diff
@@ -0,0 +1,295 @@
...
+                Param.defaultTo("host","localhost")
+                    .setDescription("the host address to use")
+            )
+            .add(
+                Param.defaultTo("port",12345)
+                    .setDescription("the designated port to connect to on the socket")
+            )
+            .add(
+                Param.defaultTo("filename","tcpserver")
+                    .setDescription("this is the name of the output file. If 'stdout', output is sent to tcpserver, not a file.")
```
jshook, line 146: This parameter should be removed from the tcpserver and tcpclient adapters. It is for stdout.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpserver/TcpServerOpDispenser.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final LongFunction<TcpServerAdapterSpace> ctxFunction;
+    private final LongFunction<String> outFunction;
+
+    public TcpServerOpDispenser(TcpServerDriverAdapter adapter, ParsedOp cmd, LongFunction<TcpServerAdapterSpace> ctxfunc) {
+        super(adapter,cmd);
+        this.ctxFunction = ctxfunc;
+        LongFunction<Object> objectFunction = cmd.getAsRequiredFunction("stmt", Object.class);
+        LongFunction<String> stringFunction = l -> objectFunction.apply(l).toString();
+        cmd.enhanceFuncOptionally(stringFunction,"suffix",String.class,(a, b) -> a+b);
```
jshook, line 34: same as above, we should remove this and let users defer to the easier "prefix{binding}suffix" format which is not supported throughout.

jshook (changes requested):
Some minor change requests, but overall looking pretty good.

adapter-tcp/src/main/java/io/nosqlbench/adapter/tcpclient/TcpClientOpDispenser.java
MikeYaacoubStax: To make sure I got this right, does this mean we no longer need to add something like:
`cmd.enhanceFuncOptionally(stringFunction,"prefix",String.class,(a, b) -> b+a);`
Because we support string templating to add a prefix to a binding, and that can do that work for us without parsing a special 'prefix' parameter?

PR #907, opened 2023-01-11
Add 'forceStop' NB CLI option in NB5 

yabinmeng:
... port over from similar functionality in NB4 (PR# [675](https://github.com/nosqlbench/nosqlbench/pull/675))

### DIFF
--- a/engine-core/src/test/java/io/nosqlbench/engine/core/ActivityExecutorTest.java
+++ b/engine-core/src/test/java/io/nosqlbench/engine/core/ActivityExecutorTest.java
@@ -65,7 +65,7 @@ public class ActivityExecutorTest {
         try {
             ad.setThreads(1);
             ae.startActivity();
-            ae.stopActivity(false);
+            ae.stopActivity();
             ae.startActivity();
             ae.startActivity();
             ExecutionResult executionResult = future.get();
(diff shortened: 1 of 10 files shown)


engine-core/src/test/java/io/nosqlbench/engine/core/ActivityExecutorTest.java
```diff
@@ -65,7 +65,7 @@ public synchronized void testRestart() {
         try {
             ad.setThreads(1);
             ae.startActivity();
-            ae.stopActivity();
+            ae.stopActivity(false);
```
jshook, line 68: I'm not a huge fan of using booleans to encode critical details like this. It's not very self explanatory.
Can we break this out into an `ae.forceStopActivity()` for example?

jshook (changes requested):
looks good in general except for the boolean flag. It would be better to have a different method here.

engine-core/src/test/java/io/nosqlbench/engine/core/ActivityExecutorTest.java
yabinmeng: I will make the changes.

yabinmeng: @jshook change have been made. A separate forceStopActivity() method is added.

jshook (approved):
LGTM!

jshook:
Thanks, @yabinmeng !

PR #908, opened 2023-01-11
Nosqlbench 904 version5

jshook:
A branch in progress for version updates


jeffbanks:
The build failure is related to [this issue](https://github.com/nosqlbench/nosqlbench/issues/913) in GHA from what I can tell.

We are going to investigate that aside from this effort.

Issue #903, opened 2023-01-11
Codeql - security advisory md5 usage


jeffbanks:
**Describe the bug**

[Scans](https://github.com/nosqlbench/nosqlbench/security/code-scanning/) are advising that MD5 digest usage in NB is replaced with stronger algo.

`MD5` is considered a broken or risky cryptographic algorithm and flagged for replacement in security advisory code scans.

**To Reproduce**
Run security code codeql against current NB main branch.

**What was Expected**
A clean security code scan where message digest conversions take place.

jeffbanks:
MD5 message warning when used, refusal old function, and removal of the actual md5 algo. Messages recommending user to utilize one of the alternatives supported (other than md5).

PR #909, opened 2023-01-12
md5 deprecation

jeffbanks:
As part of #903, removing MD5 references for high priority security advisories.


jeffbanks:
[Removal of endpoints as dead code](https://github.com/nosqlbench/nosqlbench/pull/909/commits/fbbe4f483427fa3a2d424cac46c3c1473706da30)
This is cleanup of dead code no longer utilized in nb5.

virtdata-userlibs/src/main/resources/docs-for-virtdata/bindings/funcref_conversion.md
```diff
@@ -228,6 +228,7 @@ Convert the input value to a long.
 
 
 ## ToMD5ByteBuffer
+**[ DEPRECATED ]** no longer supported use DigestToByteBuffer with stronger digest algorithm.
```
jshook, line 231: We should make the docs exporter do this for us in a separate issue.

jshook (approved):
Thanks for the thorough and defensive approach on this. Users will be happier.

virtdata-userlibs/src/main/resources/docs-for-virtdata/bindings/funcref_conversion.md
jeffbanks: I created a new docs issue to track that effort.

PR #912, opened 2023-01-12
Regular Expression Inefficiency fixes

MikeYaacoubStax:
+ Provided limited number of newlines in SSL KEY and CERT Patterns in Regex, Named the flag for more clarity
 + Corrected non-escaped minus sign in regex, and limited delimeter characters to a maximum of 2 to account for comma-space separated values
 + Limited name length to limit user input causing slow regex run, Removed starting boundary delimiter

### DIFF
--- a/nb-api/src/main/java/io/nosqlbench/api/engine/util/SSLKsFactory.java
+++ b/nb-api/src/main/java/io/nosqlbench/api/engine/util/SSLKsFactory.java
@@ -46,7 +46,8 @@ public class SSLKsFactory implements NBMapConfigurable {
     private static final SSLKsFactory instance = new SSLKsFactory();
 
     private static final Pattern CERT_PATTERN = Pattern.compile("-+BEGIN\\s+.*CERTIFICATE[^-]*-+(?:\\s|\\r|\\n){1,10}([a-z0-9+/=\\r\\n]+)-+END\\s+.*CERTIFICATE[^-]*-+", Pattern.CASE_INSENSITIVE);
-    private static final Pattern KEY_PATTERN = Pattern.compile("-+BEGIN\\s+.*PRIVATE\\s+KEY[^-]*-+(?:\\s|\\r|\\n){1,10}([a-z0-9+/=\\r\\n]+)-+END\\s+.*PRIVATE\\s+KEY[^-]*-+", Pattern.CASE_INSENSITIVE);public static final String SSL = "ssl";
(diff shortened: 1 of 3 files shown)


nb-api/src/main/java/io/nosqlbench/api/engine/util/SSLKsFactory.java
```diff
@@ -45,9 +45,8 @@ public class SSLKsFactory implements NBMapConfigurable {
 
     private static final SSLKsFactory instance = new SSLKsFactory();
 
-    private static final Pattern CERT_PATTERN = Pattern.compile("-+BEGIN\\s+.*CERTIFICATE[^-]*-+(?:\\s|\\r|\\n)+([a-z0-9+/=\\r\\n]+)-+END\\s+.*CERTIFICATE[^-]*-+", 2);
-    private static final Pattern KEY_PATTERN = Pattern.compile("-+BEGIN\\s+.*PRIVATE\\s+KEY[^-]*-+(?:\\s|\\r|\\n)+([a-z0-9+/=\\r\\n]+)-+END\\s+.*PRIVATE\\s+KEY[^-]*-+", 2);
-    public static final String SSL = "ssl";
+    private static final Pattern CERT_PATTERN = Pattern.compile("-+BEGIN\\s+.*CERTIFICATE[^-]*-+(?:\\s|\\r|\\n){1,10}([a-z0-9+/=\\r\\n]+)-+END\\s+.*CERTIFICATE[^-]*-+", Pattern.CASE_INSENSITIVE);
+    private static final Pattern KEY_PATTERN = Pattern.compile("-+BEGIN\\s+.*PRIVATE\\s+KEY[^-]*-+(?:\\s|\\r|\\n){1,10}([a-z0-9+/=\\r\\n]+)-+END\\s+.*PRIVATE\\s+KEY[^-]*-+", Pattern.CASE_INSENSITIVE);public static final String SSL = "ssl";
```
jeffbanks, line 49: Looks like a leechy line here at the end for SSL.

jeffbanks (changes requested):
Look like some good overall constraints to reduce the iterations.  One item to fix on formatting. 

Do you know if we have good test coverage for these regex patterns existing?  We may be able to run the coverage report on these to see as well, but not a show stopper IMO.

jeffbanks:
@MikeYaacoubStax Looks good!  

Thank you for resolving some high severity code scan issues.

jeffbanks (approved):

MikeYaacoubStax:
> Look like some good overall constraints to reduce the iterations. One item to fix on formatting.
> 
> Do you know if we have good test coverage for these regex patterns existing? We may be able to run the coverage report on these to see as well, but not a show stopper IMO.

@jeffbanks I can tell you that I was curious to know if these regexs had tests using them, and they all did. changing each one of them to something invalid and running the Maven tests came back with a test failure, so I believe they are being covered. However, I do not think the tests are designed to test these Regexs specifically, they must have a few specific strings to match against.

PR #928, opened 2023-01-17
Coverage report integration

jeffbanks:
Effort for integration of jacoco reports with existing dashboard.
Includes investigation of metrics generated for actual test coverage for targeted nb classfiles.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,9 +28,6 @@ jobs:
       - name: Installing dependencies
         run: mvn clean install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
 
-      - name: Verifying
-        run: mvn verify
-
       - name: Running tests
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -25,33 +25,33 @@ jobs:
...
           restore-keys: ${{ runner.os }}-m2
 
-      - name: mvn verify
-        run: mvn verify
-
       - name: Installing dependencies
         run: mvn clean install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
 
+      - name: Verifying
+        run: mvn verify
```
jshook, line 32: The install phase automatically includes verify. Is there any reason to call it again here?

jshook (commented):
One question pending... I think we might be able to remove a step.

jshook (approved):
looks good to me

PR #929, opened 2023-01-18
Some minor improvements to enable better docs

jshook:


MikeYaacoubStax (approved):

MikeYaacoubStax:
@jshook Overall looks good to me, I like the formatting weights changes. One question though, do we already have javadocs written for `cql-ring-analyzer` ,`appserver` and `summarize-values?` I see we added the Service annotation and I assume they have some javadocs that we want to dynamically generate.

jshook:
These apps should be documented as part of finishing their implementation. The reason I disabled them (via annotation) was that they were not verifiable in their current state. We can make an issues for each one to define it more clearly and make it useful before re-enabling.

PR #931, opened 2023-01-18
replace missing cancelation for shutdown hook

### LINKED ISSUE #930: Invalid error message on successful runs.
**Describe the bug**
No matter what the exit status is, the following is printed on console at the end of an nb5 session:
```
    6074 WARN  [scenarios:001] SCENARIO     Scenario was interrupted by process exit, shutting down
```

**To Reproduce**
1. Run a scenario

**What was Expected**
When there is no error to stop the scenario, the message should not appear.

jshook:
closes #930

### DIFF
--- a/engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/scenario/Scenario.java
+++ b/engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/scenario/Scenario.java
@@ -278,6 +278,7 @@ public class Scenario implements Callable<ExecutionMetricsResult> {
 
         Runtime.getRuntime().removeShutdownHook(scenarioShutdownHook);
         scenarioShutdownHook.run();
+        scenarioShutdownHook = null;
     }
 
     public void notifyException(Thread t, Throwable e) {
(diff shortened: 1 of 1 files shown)


engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/scenario/Scenario.java
```diff
@@ -278,6 +278,7 @@ private synchronized void runScenario() {
 
         Runtime.getRuntime().removeShutdownHook(scenarioShutdownHook);
         scenarioShutdownHook.run();
+        scenarioShutdownHook = null;
```
jeffbanks, line 281: I believe that the `.run()` above is essentially calling a synchronized `finish()` on this scenario instance from the scenarioShutdownHook.  If this is the case, what is the purpose of nulling out the scenarioShutdownHook here?

jeffbanks (approved):

engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/scenario/Scenario.java
jshook: This is a mutual exclusion check. The finish() method changes the volatile pointer which effectively signals whether or not the scenario was in control when it exited. The non-exceptional flow allows the Scenario it to deregister the hook and run it directly, but if this *doesn't* happen, then when the hook is run, an error is detected, as it was the side-effect of `Runtime.getRuntime().addShutdownHook(scenarioShutdownHook);` which triggered when the JVM exited. This is much like signal masking. It could probably be simplified, but the fix here was to avoid the present bug with minimal change.

PR #942, opened 2023-01-23
cleanup duplicitous activity parameters during initialization (yaml,type)

### LINKED ISSUE #937: possible tag matching regression in backward support for 5.17.0
There are some workloads in fallout integration which appear to have different tag matching behavior. This could also be an argument syntax issue, but we'll treat it as a functional regression and add coverage at least for now.

One diagnostic that is missing in 5.* which was present in 4.* is the tag filtering debug log. This will be added back in.

jshook:
This PR restores some sanity checking for activity parameters
closes #937


engine-api/src/main/java/io/nosqlbench/engine/api/activityimpl/SimpleActivity.java
```diff
@@ -652,19 +652,17 @@ protected <O> OpSequence<OpDispenser<? extends O>> createOpSequence(Function<OpT
...
             Optional<String> stmt = activityDef.getParams().getOptionalString("op", "stmt", "statement");
             Optional<String> op_yaml_loc = activityDef.getParams().getOptionalString("yaml", "workload");
             if (stmt.isPresent()) {
-                stmtsDocList = StatementsLoader.loadStmt(logger, stmt.get(), activityDef.getParams());
                 workloadSource = "commandline:" + stmt.get();
+                return StatementsLoader.loadStmt(logger, stmt.get(), activityDef.getParams());
             } else if (op_yaml_loc.isPresent()) {
-                stmtsDocList = StatementsLoader.loadPath(logger, op_yaml_loc.get(), activityDef.getParams(), "activities");
                 workloadSource = "yaml:" + op_yaml_loc.get();
+                return StatementsLoader.loadPath(logger, op_yaml_loc.get(), activityDef.getParams(), "activities");
```
jeffbanks, line 674: On line below, do we need to provide another else block to warn that neither the statement and the op yaml location was present.

engine-api/src/main/java/io/nosqlbench/engine/api/activityimpl/SimpleActivity.java
```diff
@@ -547,22 +547,33 @@ protected List<ParsedOp> loadParsedOps(NBConfiguration cfg, Optional<DriverAdapt
...
+
+                // In this case, let's try to synthesize the ops as long as at least a default driver was provided
                 if (defaultDriverAdapter.isPresent() && defaultDriverAdapter.get() instanceof SyntheticOpTemplateProvider sotp) {
                     filteredOps = sotp.getSyntheticOpTemplates(stmtsDocList, getActivityDef().getParams());
                     Objects.requireNonNull(filteredOps);
+                    if (filteredOps.size() == 0) {
+                        throw new BasicError("Attempted to create synthetic ops from driver '" + defaultDriverAdapter.get().getAdapterName() + "'" +
+                            " but no ops were created. You must provide either a workload or an op parameter. Activities require op templates.");
+                    }
+                } else { // But if there were no ops, and there was no default driver provided, we can't continue
```
jeffbanks, line 572: 👍🏼 on level of details for user, if can't continue.

jeffbanks (approved):
LGTM overall, just a few comments for consideration.

engine-api/src/main/java/io/nosqlbench/engine/api/activityimpl/motor/RunStateTally.java
```diff
@@ -107,7 +105,7 @@ public synchronized void add(RunState state) {
         counts[state.ordinal()]++;
         logger.trace(() -> this +" +"+state+ String.format(":%04d",counts[state.ordinal()]));
         if (counts[state.ordinal()]==1) {
-            logger.debug(() -> "NOTIFYing on ++-SOME edge for " + state + String.format(":%04d",counts[state.ordinal()]));
+//            logger.debug(() -> "NOTIFYing on ++-SOME edge for " + state + String.format(":%04d",counts[state.ordinal()]));
```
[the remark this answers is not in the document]
jshook: I commented out to improve debug logging, as the one uncommented one covers these logically. However I wasn't sure about completely removing them, so I punted for now.

engine-api/src/main/java/io/nosqlbench/engine/api/activityimpl/SimpleActivity.java
jshook: I left the responsibility of this to the consumer, which has more proactive checks now. There may be cases where we want to allow automation to proceed at this level. It's not clear yet.

PR #954, opened 2023-01-27
Nosqlbench 944 jsonnet, includes the jsonnet capability and some fixes

### LINKED ISSUE #944: Workload meta-templating
### Discussed in https://github.com/nosqlbench/nosqlbench/discussions/940

<div type='discussions-op-text'>

<sup>Originally posted by **jshook** January 22, 2023</sup>
In several scenarios, users have needed more expressive tools to test their systems with than just op templates. They have needed a facility to provide macro-level specification, or meta-programming of a test scenario.

These sceanrios include, for example, the ability to specify a set 1000 different operations which are structurally similar but which use a different field, or the ability to use a multitude of tables in a schema management test.

There are two primary ways of achieving this, and they are complimentary:
1. "direct API" - Expose the activity and op templating capability directly to the scenario scripting environment, so that templates can be built programmatically and then passed to the scenario as a direct data structure, rather than indirectly as a serialized data structure in a yaml or json file.
2. "pre-processor" - Allow a logical pre-processor that follows a notion of templating at a higher level, with a suitable templating language.

These can both be supported as needed for different usage scenarios. 

# Trade-offs between approaches

The direct API method is a natural extension of the existing NB runtime design. Scenario designers who are already used to building more advanced scenario scripts directly would simply have more options to pick from. Either they pass a workload file to an activity which uses the well-define op templating format (data structure), or they pass an object (data structure) as an activity parameter. In both cases, the data structure uses the well-established op template standards. With current NB machinery, this would remain an initialization-time option, as once the activity's sequence is computed, it is considered to be static for the life of the activity. This type of enhancement is already on the plan for future NB enhancements, although it does comes with trade-offs.

The pre-processor method is a way to bring less programmatic options to users. The usage of the [jsonnet](https://jsonnet.org/) templating language, for example, could allow for some very simple meta programming of op templates without requiring all users to drop down to writing code. This could allow users who are already focused on _data-driven_ and _template-oriented_ design to use a tool that is already well established in the OSS ecosystem.

# Option Three: Layering

There is another intriguing option which is _why not both?_. Given that there is merit for the use of both of these approaches, it is reasonable that bringing them both together would provide the most flexibility. The direct API method can be considered a _meta-programming_ approach and the pre-processor can be considered a _meta-templating_ approach. It is quite reasonable that someone would want to use the meta-templating _within_ their use of programmatic data structure building. 

# Enabling pre-processing

There should not be an easy way to accidentally enable jsonnet processing on data which is not intended for such. Thus, a file extension which follows jsonnet conventions should be adhered to. Specifically, any file which is referenced as `.jsonnet` should be processed, _after_ template variable expansion, into an equivalent `.json` file, and loaded with the standard workload processing logic. JSON is a subset of YAML, so this is already supported in theory.

# Enabling direct API

The case for enabling the use of direct data structures for workload data should be obvious and explicit as well. Thankfully this is easy, as it is a simple matter of what type of reference is passed from the scripting environment to the scenario controller API. Essentially, if the `workload` parameter is a JSON object, then it should be considered the equivalent data structure to anything that can be loaded from a workload YAML or JSON file.

Enabling jsonnet processing against data structure in the scripting environment is a different scenario altogether. This is the layering scenario. In this case, it should suffice to make an explicit service object in the scripting environment which provides jsonnet capability. This can simply be implemented as a scripting extension for NB. A `jsonnet` service object can provide JSON evaluation capabilities with optional context injection.

# Outcomes

If this capability is added to NB,
1. More advanced scenarios can be tested without resorting to a wholly separate layer of meta programming or scripting.
2. Advanced testing scenarios will be encoded naturally into the already portable and self-contained formats supported by NB, allowing for single-file portability between users and systems.
3. More sophisticated testing scenarios which are needed, but not presently available (whether from lack of tooling capability, or basic costs) will become more routine.

</div>

### LINKED ISSUE #951: bug: NBIO loading jsonnet files is using wrong extension in some cases

jshook:
closes #944 
closes #951


adapters-api/src/main/java/io/nosqlbench/engine/api/activityconfig/rawyaml/OpsOwner.java
```diff
@@ -20,21 +20,21 @@
...
-    private List<RawStmtDef> rawStmtDefs = new ArrayList<>();
+    private List<RawOpDef> rawOpDefs = new ArrayList<>();
 
-    public StatementsOwner() {
+    public OpsOwner() {
     }
 
-    public List<RawStmtDef> getRawStmtDefs() {
-        return rawStmtDefs;
+    public List<RawOpDef> getRawStmtDefs() {
```
jeffbanks, line 32: Do you still want this to remain Stmt or Op for the defs given it returns RawOpDef?

adapters-api/src/main/java/io/nosqlbench/engine/api/activityconfig/rawyaml/OpsOwner.java
```diff
@@ -20,21 +20,21 @@
...
+        return rawOpDefs;
     }
 
-    public void setRawStmtDefs(List<RawStmtDef> rawStmtDefs) {
-        this.rawStmtDefs = rawStmtDefs;
+    public void setRawStmtDefs(List<RawOpDef> rawOpDefs) {
```
jeffbanks, line 36: same question as the above.

jeffbanks (approved):
LGTM, just a couple of questions and comments.  The tests already in place for statements (no ops) look to provide some good coverage.  

Happy to see the jsonnet capability now, a nice addition to what we can provide.

adapters-api/src/main/java/io/nosqlbench/engine/api/activityconfig/rawyaml/OpsOwner.java
jshook: I missed this one, I'll update it.

jshook: making a separate PR which improves naming of everything

PR #958, opened 2023-01-27
Replace embedded commons3 statistics with commons-statistics 1.0

### LINKED ISSUE #957: Remove commons match included src in lieu of new commons math4 for distributions
The classes under this package path are causing issues for jacoco, and a better alternative is available now. It's time to rebase the dependency on what is available from public maven repos.

jshook:
This PR replaces an ugly hack for an urgently needed library vintage with a normal library dependency, freshly minted as 1.0. There are a few places where commons-math4 beta are fill-ins, but the core of the new code here is the modernized release version of the apache math statistics, largely based on the previous versions, just with better APIs.

closes #957 

Note to reviewers: Definitely look at each commit separately on this one to simplify the view.


virtdata-api/pom.xml
```diff
@@ -60,6 +60,28 @@
...
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-math4-legacy-exception</artifactId>
+            <version>4.0-beta1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-statistics-distribution</artifactId>
+        </dependency>
+
```
MikeYaacoubStax, line 84: These dependencies have been added to mvn-defaults, do they still need to be added here?

virtdata-lib-curves4/src/main/java/io/nosqlbench/virtdata/library/curves4/continuous/int_double/ConstantContinuous.java
```diff
@@ -32,6 +32,6 @@
 @Categories({Category.distributions})
 public class ConstantContinuous extends IntToDoubleContinuousCurve {
     public ConstantContinuous(double value, String... mods) {
-        super(new ConstantContinuousDistribution(value), mods);
```
MikeYaacoubStax, line 35: I can't find the docs on ConstantContinuousDistribution anymore:
[](https://commons.apache.org/proper/commons-statistics/commons-statistics-distribution/apidocs/org/apache/commons/statistics/distribution/ConstantContinuousDistribution.html)
Are we sure an EmpiricalDistribution will fullfill the same role?

MikeYaacoubStax, line 35: I can't find the docs on ConstantContinuousDistribution anymore on apache commons:
https://commons.apache.org/proper/commons-statistics/commons-statistics-distribution/apidocs/org/apache/commons/statistics/distribution/ConstantContinuousDistribution.html
Are we sure the EmpiricalDistribution can execute the same role?

virtdata-api/pom.xml
jshook: The mvn-default entry is for "dependency management" and only serves to set the version in a central place. (Thus why this doesn't have to have one)

virtdata-lib-curves4/src/main/java/io/nosqlbench/virtdata/library/curves4/continuous/int_double/ConstantContinuous.java
jshook: The implementation of the current form is effectively hidden in a private implementation behind the EmpiricalDistribution, so we are just using a different way to get it.

PR #961, opened 2023-01-27
This consolidates our jacoco into mvn-default

### LINKED ISSUE #955: consolidate jacoco maven mojo
The initial jacoco CI/CD tooling works in most cases, but there are some corner cases where it breaks. The config needs to be centralized into the mvn-defaults pom so that module names references are not needed within the container pom. Also, reports need to use wildcards to match within each module. If we want to gen reports and such from any aggregates, we can put those actions into the most dependent (last to build) module, which is nbr-examples.

jshook:
There will be additional work to make this do exactly what we want, but with this structure, we can avoid having to maintain references.

closes #955


mvn-defaults/pom.xml
```diff
@@ -484,7 +487,118 @@
                 <version>1.7.1</version>
             </extension>
         </extensions>
+
         <plugins>
+
```
jeffbanks, line 492: commented out here.

jeffbanks (approved):
Just the one comment on the dead code in default pom for jacoco.  LGTM.  Nice reduction of configuration in other POMs.  Good to know that works this way w/ the new Maven configuration.

mvn-defaults/pom.xml
jshook: I'll remove this to tidy. If we want the code back to build on, it's in our git log.

PR #971, opened 2023-02-01
minimum fix needed for example in docs

### LINKED ISSUE #969: minor workload fixes for docs updates
Some of the classic workloads used in the docs need to be updated so it can provide a better experience for new users.

jshook:
This fixes a binding and some tag structure so that we can use it in the docs

closes #969


jeffbanks (approved):
LGTM

PR #973, opened 2023-02-01
Adds missing logic to accomodate the Errored state

### LINKED ISSUE #972: Include Errored state into activity threading model
Some event handling logic is over-strict for the recently added Errored state.
These need to be relaxed some what for shutdown so that they don't incorrectly throw an error that users can't do something with.

jshook:
closes #972

### DIFF
--- a/engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/activity/ActivityExecutor.java
+++ b/engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/activity/ActivityExecutor.java
@@ -299,7 +299,9 @@ public class ActivityExecutor implements ActivityController, ParameterMap.Listen
                 break;
             case Finished:
             case Stopping:
-                throw new RuntimeException("Invalid requested state in activity executor:" + activity.getRunState());
+            case Errored:
+                break;
+//                throw new RuntimeException("Invalid requested state in activity executor:" + activity.getRunState());
 
             default:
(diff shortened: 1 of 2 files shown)


engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/activity/ActivityExecutor.java
```diff
@@ -314,16 +316,16 @@ private void awaitAlignmentOfMotorStateToActivityState() {
             case Running:
                 tally.awaitNoneOther(RunState.Running, RunState.Finished);
                 break;
+            case Errored:
+            case Stopping:
             case Stopped:
-                tally.awaitNoneOther(RunState.Stopped, RunState.Finished);
+                tally.awaitNoneOther(RunState.Stopped, RunState.Finished, RunState.Errored);
```
jeffbanks, line 322: This may help for what I've been seeing on local runs!

jeffbanks (approved):
LGTM, awaitAlignmentOfMotorStateToActivityState changes may help with local scenario runs found recently.

engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/activity/ActivityExecutor.java
jshook: yes!

PR #988, opened 2023-02-03
fixes for missing errored state in activity

### LINKED ISSUE #987: Actions thread state model is missing valid transition from stop*->Errored
In the case that an activity has a short run, or a low number of cycles *and* there is an error during it's shutdown or cycle execution, it will transition by default to Stopping and then Stopped via normal flow, but the async exception handling logic may not flag an error on the activity early enough.

When there is an error, the final state of the activity should be Errored in any case.
When there isn't an error, the final state should never be Errored. (Should be either Finished or Stopped).

This needs further review, but the first fix to be applied should be to allow Stopping  -> Errored. Both the transition and final states Stopping and Stopped should be allowed to be moved to Errored, because of a potential race. But, a better solution is to make sure that all transitions are atomic and that we account for each transition uniquely.

jshook:
closes #987


jeffbanks (approved):
LGTM

Issue #1001, opened 2023-02-06
BUG: NumberFormatException during activity initialization


jshook:
The thread pool for activities was being initialized before the activity configuration was settled. This results in a race condition between motor threads and the activity init logic which flattens symbolic values into numeric values. The motor threads are not expecting to read symbolic values, so this results in an NFE.

PR #1002, opened 2023-02-06
fix race condition between activity init and threadpool start

jshook:
This fixes the actual issue which was causing the NumberFormatException.
I added some smaller improvements which helped with diagnosis and testing.
(commits are not itemized on this one)

closes: #1001


jeffbanks (commented):
Just a few comments, good stuff here!

PR #1005, opened 2023-02-06
Simplify and optimize Pulsar driver schema type mapping

lhotari:
### Motivation

PulsarActivityUtil contains a lot of unnecessary logic for mapping a Pulsar SchemaType name to a Schema instance. 
This could be replaced with the usage of Pulsar client's functionality. Besides being verbose, this could be considered as not very optimal from performance perspective, although this part of the code hasn't been a performance bottleneck.

One part that should be cached is the Avro schema parsing. I don't have profiling results, but it's fairly obvious that a performance tool shouldn't be adding overhead by reading a schema file from disk, possibly in every cycle.

### Modifications

- simplity Pulsar SchemaType name mapping to Schema instance
- add caching to Avro schema parsing
- Use UTF8 encoding for reading Avro schema files. It was US_ASCII, which doesn't make sense.


eolivelli (approved):
Looks good

jshook (approved):
LGTM

yabinmeng:
Shouldn't we fix this in the main branch?

lhotari:
> Shouldn't we fix this in the main branch?

@yabinmeng yes, the PR is #1015

PR #1009, opened 2023-02-07
remove double append of markdown prefix

### LINKED ISSUE #1007: bundled app export-docs is sending extra headers from blurbs and frontmatter
There shouldn't be two copies of the heading section in the exported-docs.zip file.

```
---
weight: 20
title: datetime functions
---

Functions in this category know about times and dates, datetimes, seconds or millisecond epoch times, and so forth.

Some of the functions in this category are designed to allow testing of UUID types which are usually designed to avoid
determinism. This makes it possible to test systems which depend on UUIDs but which require determinism in test data.
This is strictly for testing use. Breaking the universally-unique properties of UUIDs in production systems is a bad
idea. Yet, in testing, this determinism is quite useful.

---
title: datetime functions
weight: 20
---

Functions in this category know about times and dates, datetimes, seconds or millisecond epoch times, and so forth.

```

jshook:
closes #1007
This  was happening because there are multiple paths which are matching, both logical and physical.
Rather than reduce the search paths, they now simply continue to the next when the first match is found for a category.

### DIFF
--- a/nbr/src/main/java/io/nosqlbench/api/docsapi/docexporter/BundledMarkdownZipExporter.java
+++ b/nbr/src/main/java/io/nosqlbench/api/docsapi/docexporter/BundledMarkdownZipExporter.java
@@ -33,9 +33,6 @@ import java.nio.file.StandardOpenOption;
 import java.util.Date;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


msmygit (approved):
LGTM 👍🏼

PR #1014, opened 2023-02-07
Upgrade maven plugins and remove versions of plugins where version is defined in pluginManagement

lhotari:
### Motivation

I happened to get an error `Cannot resolve plugin org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M6`.
I know from experience that 3.0.0-M6 version has caused issues in the Pulsar build. That's why I checked the used plugin versions. I noticed that there's inconsistency in the way plugin versions are managed. 
When plugin versions are defined in `pluginManagement`, the versions should be omitted in the actual `plugins` section.
Since I was at it, I think it's better to upgrade the plugin versions to the latest released versions since there was a mixture of old and new plugin versions.

### Modifications

- upgrade maven plugin versions to latest released versions
- remove versions from `plugins` for those plugin versions that are defined in `pluginManagement`

### DIFF
--- a/mvn-defaults/pom.xml
+++ b/mvn-defaults/pom.xml
@@ -501,7 +501,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-resources-plugin</artifactId>
-                <version>3.2.0</version>
                 <configuration>
                     <propertiesEncoding>ISO-8859-1</propertiesEncoding>
                 </configuration>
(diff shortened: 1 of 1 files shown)


eolivelli (approved):
LGTM

jshook (approved):
LGTM, Thanks!

jshook:
@all-contributors please add @lhotari for ci/cd and build improvements

jshook:
@all-contributors please add @lhotari for code, maintenance, platform

PR #983, opened 2023-02-03
Optimize major bottleneck in the NB Pulsar driver when running a consumer workload

lhotari:
We performed some profiling with Java Flight Recorder for a Nosqlbench Pulsar workload since the CPU consumption and GC collection time was high. 

Here are some screenshots of the Flamegraph:

image

image

This PR targets this bottleneck.

Caches should be using `computeIfAbsent` to prevent races. It was previously used in one location in `PulsarSpace` class. This PR changes also other code locations in `PulsarSpace` to use it for lookups to `producers`, `consumers` and `readers` caches. The Apache Commons Lang3 Pair and Triple classes are used as cache keys to reduce the number of generated String objects. The previous cache key solution was inefficient.

Here's an example of an inefficient method:
```
    public static boolean isValidSubscriptionType(String item) {
        return Arrays.stream(SUBSCRIPTION_TYPE.values()).anyMatch(t -> t.label.equals(item));
    }
```

It's recommended to create a static field in the enum to optimize such mappings
```java
    public enum SUBSCRIPTION_TYPE {
        Exclusive("Exclusive"),
        Failover("Failover"),
        Shared("Shared"),
        Key_Shared("Key_Shared");

        public final String label;

        SUBSCRIPTION_TYPE(String label) {
            this.label = label;
        }

        private static final Set<String> LABELS = Stream.of(values()).map(v -> v.label)
            .collect(Collectors.toUnmodifiableSet());

        public static boolean isValidLabel(String label) {
            return LABELS.contains(label);
        }
    }

    public static boolean isValidSubscriptionType(String item) {
        return SUBSCRIPTION_TYPE.isValidLabel(item);
    }
```

The same approach works for more complex solutions, for example

```java
    public enum SEQ_ERROR_SIMU_TYPE {
        OutOfOrder("out_of_order"),
        MsgLoss("msg_loss"),
        MsgDup("msg_dup");

        public final String label;

        SEQ_ERROR_SIMU_TYPE(String label) {
            this.label = label;
        }

        private static final Map<String, SEQ_ERROR_SIMU_TYPE> MAPPING = Stream.of(values())
            .flatMap(simuType ->
                Stream.of(simuType.label,
                        simuType.label.toLowerCase(),
                        simuType.label.toUpperCase(),
                        simuType.name(),
                        simuType.name().toLowerCase(),
                        simuType.name().toUpperCase())
                    .distinct().map(key -> Map.entry(key, simuType)))
[...]

PR #1015, opened 2023-02-07
Optimize major performance bottleneck in Pulsar adapter

lhotari:
### Motivation

See #983 and #1005 which targeted the nb4-maintenance branch. This PR ports similar changes to the main branch.

### Modifications

* String operations with enums can become a bottleneck with high throughput. 
  * The optimization is to pre-compute useful data structures when the enum is loaded.
* Caches should be using `computeIfAbsent` to prevent races. 
  * Use an efficient key for the caches. In Java 17, using a `record` as a key is the simplest solution.
* simplity Pulsar SchemaType name mapping to Schema instance
* add caching to Avro schema parsing
* Use UTF8 encoding for reading Avro schema files. It was US_ASCII, which doesn't make sense.

### DIFF
--- a/adapter-pulsar/src/main/java/io/nosqlbench/adapter/pulsar/dispensers/PulsarBaseOpDispenser.java
+++ b/adapter-pulsar/src/main/java/io/nosqlbench/adapter/pulsar/dispensers/PulsarBaseOpDispenser.java
@@ -37,7 +37,6 @@ import java.util.*;
 import java.util.function.LongFunction;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
-import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 
 public abstract  class PulsarBaseOpDispenser extends BaseOpDispenser<PulsarOp, PulsarSpace> implements NBNamedElement {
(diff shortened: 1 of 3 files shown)


adapter-pulsar/src/main/java/io/nosqlbench/adapter/pulsar/dispensers/PulsarBaseOpDispenser.java
```diff
@@ -558,17 +517,12 @@ public Reader<?> getReader(String cycleTopicName,
                 //    startMsgId = MessageId.latest;
                 //}
 
-                reader = readerBuilder.startMessageId(startMsgId).create();
-
+                return readerBuilder.startMessageId(startMsgId).create();
             } catch (PulsarClientException ple) {
                 ple.printStackTrace();
```
eolivelli, line 522: what about taking this chance of changing this code to remove "printStackTrace" and use the logger (and also pass the exception as the cause of the RuntimeException below ?

eolivelli (approved):
Perfect

jshook:
@all-contributors please add @lhotari for pull requests, bug fixes, performance optimization, and code reviews

Issue #1019, opened 2023-02-07
Scenario updates - Part II


jeffbanks:
**Describe the bug**
An additional round of changes required for baseline scenarios to ensure proper alignment with workload model changes introduced in `nb5`.

**To Reproduce**
There are scenarios existing that are not working as expected.

**What was Expected**
Scenarios should run with nb5.

jeffbanks:
Changes pushed for release target `v5.17.1`.

Includes `cql-starter` scenario for learning purposes.

PR #1031, opened 2023-02-07
Phase 2 scenario upgrades #1019

jeffbanks:


jshook (approved):
LGTM, rolling these in

PR #1080, opened 2023-02-10
remove wasteful IO polling in NBIO

jshook:
The `.one()` method will now look directly for the pathnames, non-prefixed first, with suffixes added when provided.

This avoids using a scanning method across all possible matching names unless absolutely needed.


MikeYaacoubStax (approved):
All Good to Go!

jshook:
@MikeYaacoubStax TY, I fixed the build errors, but they were trivial test fixes.

PR #1102, opened 2023-02-14
Nosqlbench 994 unused driver warning

jshook:
a few small issues in one PR


jeffbanks (approved):
Looks good, thanks for adding Discard() option!

PR #1114, opened 2023-02-15
Adapter and ops loader fixes

jeffbanks:


adapter-cqld4/src/main/resources/activities/baselines/cql-iot-dse.yaml
```diff
@@ -43,7 +43,7 @@ blocks:
         };
 
       truncate-table: |
-        truncate table <<keyspace:baselines>>.<<table:iot>>;
+        truncate table TEMPLATE(keyspace, baselines).TEMPLATE(table:iot);
```
jshook, line 46: The space after the comma is significant here, although in this position it should not be an issue.

PR #1117, opened 2023-02-17
Http Adapter Fixes for Stargate

jeffbanks:
Includes:
* Formatting fixes
* Token externalized
* Minimal cycles introduced for defaults
  * this is a new pattern being applied to built-in scenarios to avoid confusion with initial usage.
* Standardize for nb5

### DIFF
--- a/adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
+++ b/adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
@@ -21,7 +21,7 @@ bindings:
   weighted_hosts: WeightedStrings('<<restapi_host:stargate>>')
 
   request_id: ToHashedUUID(); ToString();
-  request_token: ToString(); TextOfFile(TEMPLATE(stargate_tokenfile,'data/stargate_token.txt')));
+  request_token: ToString(); TextOfFile(TEMPLATE(stargate_tokenfile,'data/stargate_token.txt'));
 
   seq_key: Mod(<<keycount:10000000>>); ToString() -> String
   seq_value: Hash(); Mod(<<valuecount:1000000000>>); ToString() -> String
--- a/adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
+++ b/adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
@@ -9,8 +9,8 @@ description: |
 scenarios:
   default:
     schema: run driver=http tags==block:schema threads==1 cycles==UNDEF
-    rampup: run driver=http tags==block:"rampup.*" cycles===TEMPLATE(rampup-cycles,10000000) threads=auto
-    main: run driver=http tags==block:"main.*" cycles===TEMPLATE(main-cycles,10000000) threads=auto
+    rampup: run driver=http tags==block:"rampup.*" cycles===TEMPLATE(rampup-cycles,10) threads=auto
+    main: run driver=http tags==block:"main.*" cycles===TEMPLATE(main-cycles,10) threads=auto
 
 bindings:
   # To enable an optional weighted set of hosts in place of a load balancer
@@ -21,7 +21,7 @@ bindings:
   weighted_hosts: WeightedStrings('<<restapi_host:stargate>>')
 
   request_id: ToHashedUUID(); ToString();
-  request_token: ToString(); TextOfFile(TEMPLATE(stargate_tokenfile,'data/stargate_token.txt'));
+  request_token: ToString(); TextOfFile("TEMPLATE(stargate_tokenfile,data/stargate_token.txt)");
 
   seq_key: Mod(<<keycount:10000000>>); ToString() -> String
   seq_value: Hash(); Mod(<<valuecount:1000000000>>); ToString() -> String
--- a/adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
+++ b/adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
@@ -21,7 +21,7 @@ bindings:
   weighted_hosts: WeightedStrings('<<restapi_host:stargate>>')
 
   request_id: ToHashedUUID(); ToString();
-  request_token: ToString(); TextOfFile("TEMPLATE(stargate_tokenfile,data/stargate_token.txt)");
+  request_token: ToString(); FirstLines("TEMPLATE(stargate_tokenfile,data/stargate_token.txt)");
 
   seq_key: Mod(<<keycount:10000000>>); ToString() -> String
   seq_value: Hash(); Mod(<<valuecount:1000000000>>); ToString() -> String
--- a/adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
+++ b/adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
@@ -21,7 +21,7 @@ bindings:
   weighted_hosts: WeightedStrings('<<restapi_host:stargate>>')
 
   request_id: ToHashedUUID(); ToString();
-  request_token: ToString(); TextOfFile("TEMPLATE(stargate_tokenfile,data/stargate_token.txt)")
+  request_token: ToString(); TextOfFile(TEMPLATE(stargate_tokenfile,'data/stargate_token.txt')));
 
   seq_key: Mod(<<keycount:10000000>>); ToString() -> String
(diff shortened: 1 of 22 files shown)


adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
```diff
@@ -21,7 +21,7 @@ bindings:
   weighted_hosts: WeightedStrings('<<restapi_host:stargate>>')
 
   request_id: ToHashedUUID(); ToString();
-  request_token: ToString(); TextOfFile("TEMPLATE(stargate_tokenfile,data/stargate_token.txt)")
+  request_token: ToString(); TextOfFile(TEMPLATE(stargate_tokenfile,'data/stargate_token.txt'));
```
jshook, line 24: The quotes on this are part of the value, but should probably be outside the TEMPLATE macro.

adapter-http/src/main/resources/activities/baselines/http-rest-timeseries.yaml
```diff
@@ -27,10 +26,10 @@ bindings:
 
   # http request id
   request_id: ToHashedUUID(); ToString();
-  request_token: ToString(); TextOfFile("TEMPLATE(stargate_tokenfile,data/stargate_token.txt)")
+  request_token: ToString(); TextOfFile(TEMPLATE(stargate_tokenfile,'data/stargate_token.txt'));
```
jshook, line 29: quote placement (move outside of TEMPLATE)

adapter-http/src/main/resources/activities/documents-api/http-docsapi-crud-dataset.yaml
```diff
@@ -7,11 +7,12 @@ description: |
...
 scenarios:
   default:
-    schema:   run driver=http tags==block:"schema.*" threads==1 cycles==UNDEF
-    write:    run driver=http tags==name:'write.*' cycles===TEMPLATE(write-cycles,TEMPLATE(docscount,10000000)) threads=auto errors=timer,warn
-    read:     run driver=http tags==name:'read.*' cycles===TEMPLATE(read-cycles,TEMPLATE(docscount,10000000)) threads=auto errors=timer,warn
-    update:   run driver=http tags==name:'update.*' cycles===TEMPLATE(update-cycles,TEMPLATE(docscount,10000000)) threads=auto errors=timer,warn
-    delete:   run driver=http tags==name:'delete.*' cycles===TEMPLATE(delete-cycles,TEMPLATE(docscount,10000000)) threads=auto errors=timer,warn
+    schema: run driver=http tags==block:"schema.*" threads==1 cycles==UNDEF
+    main: run driver=http tags==block:"main.*" cycles===TEMPLATE(main-cycles,TEMPLATE(docscount,10)) threads=auto errors=timer,warn
+    # write:    run driver=http tags==block:"write.*" cycles===TEMPLATE(write-cycles,TEMPLATE(docscount,10)) threads=auto errors=timer,warn
```
jshook, line 12: Are the supposed to be commented out?

adapter-http/src/main/resources/activities/documents-api/http-docsapi-crud-dataset.yaml
```diff
@@ -22,9 +23,11 @@ bindings:
   weighted_hosts: WeightedStrings('<<docsapi_host:stargate>>')
   # http request id
   request_id: ToHashedUUID(); ToString();
+  request_token: Discard(); TextOfFile(TEMPLATE(stargate_tokenfile,'data/stargate_token.txt'));
 
   seq_key: Mod(<<docscount:10000000>>); ToString() -> String
   random_key: Uniform(0,<<docscount:10000000>>); ToString() -> String
+  document_json: Discard(); TextOfFile('data/docsapi-dataset.json', false));
```
jshook, line 30: I would prefer to see us split this out to a dedicated function, like `File(...)`. Booleans as signatures are not direct in their meaning, and often confuse users.

adapter-http/src/main/resources/data/stargate_token.txt
```diff
@@ -1 +1 @@
-# <<put-token-here>>
\ No newline at end of file
+<<your-token-here>>
```
jshook, line 1: We should probably move this out of the resources, since having a "file not found" error is much more desirable than having a "invalid credentials" issue in the event that they don't have a filesystem version of this file. At least, we should make the names distinctive where this is referenced in basic workloads. Comments? Ideas?

jshook (changes requested):
Good changes in the workloads, only minor changes requested.

adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
jeffbanks: Added.

adapter-http/src/main/resources/activities/baselines/http-rest-timeseries.yaml
jeffbanks: Added.

adapter-http/src/main/resources/data/stargate_token.txt
jeffbanks: 👍🏼 to removing this file entirely.

adapter-http/src/main/resources/activities/documents-api/http-docsapi-crud-dataset.yaml
jeffbanks: Right, if this is the case to include File(...), we may need to consider the existing `TextOfFile` to be something more specific by name (e.g. LineOfFile(...) to indicated that only one line is being returned.  This will require some refactoring throughout its current usage in the scenarios and tests but when I read TextOfFile I'm thinking basically what we would now be introducing and intending with File(...).  wdyt?

jeffbanks: Changed to FirstLines(<path>, <number>) where no number specified indicates 1 line.

adapter-http/src/main/resources/activities/documents-api/http-docsapi-crud-dataset.yaml
jeffbanks: Yes, leaving for historical reference as those don't appear to be used at the moment.

jeffbanks:
@jshook this one is good for second review.

adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
```diff
@@ -21,7 +21,7 @@ bindings:
   weighted_hosts: WeightedStrings('<<restapi_host:stargate>>')
 
   request_id: ToHashedUUID(); ToString();
-  request_token: ToString(); TextOfFile("TEMPLATE(stargate_tokenfile,data/stargate_token.txt)")
+  request_token: ToString(); FirstLines("TEMPLATE(stargate_tokenfile,data/stargate_token.txt)");
```
jeffreyscarpenter, line 24: So you have to write a token to a file in order to run this test? Should that be documented in the accompanying markdown file? I prefer the way things used to work where you pass the token as a property when invoking the script.

jeffreyscarpenter (commented):
Just had one comment regarding handling of tokens.

adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
jeffbanks: @jeffreyscarpenter I agree, thanks for pointing this out.  The original intent was to perform  something similar to:  

`TEMPLATE(stargate_token, FirstLines(data/stargate_token.txt))`

I'll need to adjust some things to allow for this passing of the token with the externalized file option.  Good catch!

Also, as context, in the next iteration of the http activities (after we shake out issues) I was planning to obtain the token via http for the user to avoid having to supply a token at all.  However, it would still provide a direct submission as well.  wdyt?

I'll get the initial issue adjusted for this upcoming release.

jshook: noted, merging this for preview release

jshook (approved):
LGTM, Thanks. Merging this in for 5.17.2-preview

adapter-http/src/main/resources/activities/baselines/http-rest-keyvalue.yaml
jeffreyscarpenter: an option to get the token via http would be useful.

PR #1127, opened 2023-02-22
[kafka-adapter] Add e2e latency metrics

MMirelli:
Currently, the kafka adapter doensn't support neither e2e latency metrics nor e2e error metrics. These are all defined in the pulsar adapter: https://github.com/nosqlbench/nosqlbench/blob/12f8697e0e27754074eca413451c0707d0663d70/adapter-pulsar/src/main/java/io/nosqlbench/adapter/pulsar/util/PulsarAdapterMetrics.java#L42-L59.

This aims at integrating the e2e latency metrics.

### DIFF
--- a/adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/dispensers/KafkaBaseOpDispenser.java
+++ b/adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/dispensers/KafkaBaseOpDispenser.java
@@ -37,7 +37,7 @@ import java.util.function.Predicate;
 
 public abstract  class KafkaBaseOpDispenser extends BaseOpDispenser<KafkaOp, KafkaSpace> implements NBNamedElement {
 
-    private final static Logger logger = LogManager.getLogger("PulsarBaseOpDispenser");
+    private final static Logger logger = LogManager.getLogger("KafkaBaseOpDispenser");
 
     protected final ParsedOp parsedOp;
     protected final KafkaAdapterMetrics kafkaAdapterMetrics;
@@ -135,6 +135,6 @@ public abstract  class KafkaBaseOpDispenser extends BaseOpDispenser<KafkaOp, Kaf
 
     @Override
     public String getName() {
-        return "MessageConsumerOpDispenser";
+        return "KafkaBaseOpDispenser";
     }
 }
--- a/adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/dispensers/KafkaBaseOpDispenser.java
+++ b/adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/dispensers/KafkaBaseOpDispenser.java
@@ -22,11 +22,11 @@ import io.nosqlbench.adapter.kafka.exception.KafkaAdapterInvalidParamException;
 import io.nosqlbench.adapter.kafka.ops.KafkaOp;
 import io.nosqlbench.adapter.kafka.util.KafkaAdapterMetrics;
 import io.nosqlbench.adapter.kafka.util.KafkaAdapterUtil;
+import io.nosqlbench.api.config.NBNamedElement;
 import io.nosqlbench.engine.api.activityimpl.BaseOpDispenser;
 import io.nosqlbench.engine.api.activityimpl.uniform.DriverAdapter;
 import io.nosqlbench.engine.api.templating.ParsedOp;
 import org.apache.commons.lang3.BooleanUtils;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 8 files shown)


adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaConsumer.java
```diff
@@ -157,15 +167,30 @@ public void onComplete(Map<TopicPartition, OffsetAndMetadata> map, Exception e)
...
                 }
             }
         }
     }
 
+    private void updateE2ELatencyMetric(ConsumerRecord<String, String> record) {
+        long startTimeStamp = 0L;
+        switch (e2eStartingTimeSrc) {
+            case MESSAGE_PUBLISH_TIME:
+                startTimeStamp = record.timestamp();
```
eolivelli, line 185: how is r`ecord.timestamp` populated ?

adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/util/EndToEndStartingTimeSource.java
```diff
@@ -0,0 +1,23 @@
+package io.nosqlbench.adapter.kafka.util;
+
+/*
+ * Copyright (c) 2022 nosqlbench
```
msmygit, line 4:
```suggestion
 * Copyright (c) 2022-2023 nosqlbench
```

adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaConsumer.java
yabinmeng: From the Kafka API Doc. The NB Kafka producer uses the default setting and doesn't explicitly set a timestamp.

```
The record also has an associated timestamp. If the user did not provide a timestamp, the producer will stamp the record with its current time. The timestamp eventually used by Kafka depends on the timestamp type configured for the topic.

If the topic is configured to use CreateTime, the timestamp in the producer record will be used by the broker. 
```

yabinmeng (changes requested):
LGTM except 2 minor naming changes

yabinmeng:
@MMirelli regarding your comment about rate limit, right now it is very hard to do rate limiting especially on the consumer side. This is because in Kafka, consumer receive messages in batches. This means one NB cycle for Kafka message receiving may receive 0 message or multiple messages. Most likely what happens is for a few NB cycles, there is no message received and then for the next few NB cycles, there are multiple messages received in each cycle. The NB cycle rate limiting won't exactly reflect the desired message receiving rate limiting.


`Please let me know if rate throttling can be enabled in some other way. This is where I learnt cyclerate is the main method: https://github.com/nosqlbench/nosqlbench/blob/1711c5c02c78017c0081041ac80fbc5868de37c3/engine-docs/src/main/resources/docs-for-nb/getting_started/01_example_commands.md#run-the-main-test-phase.`

adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaConsumer.java
MMirelli: Here is the Kafka API Doc, referenced by Yabin: https://kafka.apache.org/34/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html.  I will make sure to configure the topic with `CreateTime`.

adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/dispensers/KafkaBaseOpDispenser.java
```diff
@@ -132,4 +132,9 @@ protected LongFunction<String> lookupMandtoryStrOpValueFunc(String paramName) {
 
         return stringLongFunction;
     }
+
+    @Override
+    public String getName() {
+        return "MessageConsumerOpDispenser";
```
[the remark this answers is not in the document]
MMirelli: good catch, thank you!

adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaConsumer.java
MMirelli: It seems this can be done by adding `message.timestamp.type: CreateTime` to the kafka_config.properties. But, to my understanding, that file is not used. At least nothing seems to be using the result stored in this variable https://github.com/nosqlbench/nosqlbench/blob/e832e5a5897978d1b46e01bc5e17b29e6f61f5ba/adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/dispensers/KafkaBaseOpDispenser.java#L79. 

Anyway according to [the doc](https://docs.confluent.io/platform/current/installation/configuration/topic-configs.html#message-timestamp-type) the default is `CreateTime`. So, we should be good.

PR #1129, opened 2023-02-22
nosqlbench-1092-parsedop-captures

### LINKED ISSUE #1092: Expose parsed capture points on ParsedOp API
The parser for capture points is already implemented, but it needs to be brought up to the ParsedOp level.
This will initially support only string form operations, but will be expanded to support capture points in structured op templates in a separate issue.

jshook:
closes #1092


MikeYaacoubStax (approved):
All good!

PR #1135, opened 2023-02-24
[kafka-adapter] Add e2e error metrics

MMirelli:
Currently, the kafka adapter doensn't support neither e2e latency metrics nor e2e error metrics. These are all defined in the pulsar adapter: https://github.com/nosqlbench/nosqlbench/blob/12f8697e0e27754074eca413451c0707d0663d70/adapter-pulsar/src/main/java/io/nosqlbench/adapter/pulsar/util/PulsarAdapterMetrics.java#L42-L59.

This aims at integrating the e2e error metrics.

PR #1143, opened 2023-03-01
[kafka-adapter] Add e2e error metrics -- to be continued

MMirelli:
Currently, the kafka adapter doensn't support neither e2e latency metrics nor e2e error metrics. These are all defined in the pulsar adapter: https://github.com/nosqlbench/nosqlbench/blob/12f8697e0e27754074eca413451c0707d0663d70/adapter-pulsar/src/main/java/io/nosqlbench/adapter/pulsar/util/PulsarAdapterMetrics.java#L42-L59.
This aims at integrating the e2e error metrics.

This contains the continuation of https://github.com/nosqlbench/nosqlbench/pull/1135.

### DIFF
--- a/adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaProducer.java
+++ b/adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaProducer.java
@@ -20,6 +20,8 @@ package io.nosqlbench.adapter.kafka.ops;
 import io.nosqlbench.adapter.kafka.KafkaSpace;
 import io.nosqlbench.adapter.kafka.exception.KafkaAdapterUnexpectedException;
 import io.nosqlbench.adapter.kafka.util.KafkaAdapterUtil;
+import io.nosqlbench.adapter.pulsar.util.MessageSequenceNumberSendingHandler;
+import io.nosqlbench.adapter.pulsar.util.PulsarAdapterUtil;
 import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
@@ -31,6 +33,9 @@ import org.apache.kafka.common.errors.ProducerFencedException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.util.Map;
+import java.util.HashMap;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import org.apache.kafka.common.errors.TimeoutException;
@@ -45,6 +50,10 @@ public class OpTimeTrackKafkaProducer extends OpTimeTrackKafkaClient {
     private final boolean asyncMsgAck;
     private final boolean transactEnabledConfig;
     private final int txnBatchNum;
+    private ThreadLocal<Map<String, MessageSequenceNumberSendingHandler>> MessageSequenceNumberSendingHandlersThreadLocal =
+        ThreadLocal.withInitial(HashMap::new);
+    private final boolean seqTracking;
+    private final Set<PulsarAdapterUtil.MSG_SEQ_ERROR_SIMU_TYPE> errSimuTypeSet;
 
     enum TxnProcResult {
         SUCCESS,
@@ -67,11 +76,15 @@ public class OpTimeTrackKafkaProducer extends OpTimeTrackKafkaClient {
                                     boolean asyncMsgAck,
                                     boolean transactEnabledConfig,
                                     int txnBatchNum,
+                                    boolean seqTracking,
+                                    Set<PulsarAdapterUtil.MSG_SEQ_ERROR_SIMU_TYPE> errSimuTypeSet,
                                     KafkaProducer<String, String> producer) {
         super(kafkaSpace);
         this.asyncMsgAck = asyncMsgAck;
         this.transactEnabledConfig = transactEnabledConfig;
         this.txnBatchNum = txnBatchNum;
+        this.seqTracking = seqTracking;
+        this.errSimuTypeSet = errSimuTypeSet;
         this.transactionEnabled = transactEnabledConfig && (txnBatchNum > 2);
         this.producer = producer;
     }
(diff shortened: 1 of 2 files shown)


adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaConsumer.java
```diff
@@ -191,6 +209,16 @@ private void updateE2ELatencyMetric(ConsumerRecord<String, String> record) {
         }
     }
 
+    private void checkAndUpdateMessageErrorCounter(ConsumerRecord<String, String> record) {
+        String msgSeqIdStr = new String(record.headers().lastHeader(PulsarAdapterUtil.MSG_SEQUENCE_NUMBER).value());
```
eolivelli, line 213: what happens if there is no such header ?
do we see here a null or does it throw an exception ?

adapter-kafka/pom.xml
```diff
@@ -76,6 +76,12 @@
             <artifactId>commons-configuration2</artifactId>
             <version>2.8.0</version>
         </dependency>
+        <dependency>
+            <groupId>io.nosqlbench</groupId>
+            <artifactId>adapter-pulsar</artifactId>
+            <version>5.17.2-SNAPSHOT</version>
```
eolivelli, line 82: this should be ${project.version)

adapter-kafka/pom.xml
```diff
@@ -76,6 +76,12 @@
             <artifactId>commons-configuration2</artifactId>
             <version>2.8.0</version>
         </dependency>
+        <dependency>
+            <groupId>io.nosqlbench</groupId>
```
eolivelli, line 80: ${project.groupId}

adapter-kafka/pom.xml
```diff
@@ -76,6 +76,12 @@
...
+            <artifactId>adapter-pulsar</artifactId>
```
eolivelli, line 81: please add a comment that explains why we need this dependency

eolivelli (commented):
overall LGTM
I have left a couple of comments

MMirelli:
In order to reuse the main util tracking the end-to-end errors (https://github.com/nosqlbench/nosqlbench/pull/1143/files#diff-80f671d2680a92638d696e350f048425a61a168840f63bf698dbcdb42e39141bR212-R221) I had to add `pulsar-adapter` as a dependency of the kafka adapter: https://github.com/nosqlbench/nosqlbench/pull/1143/files#diff-f6388beadc327908ba7cf42521a8083a1697d956f144edf7f25d1d4a4047a9c5R79-R84. 

In general, I think this is a bad practice. @jshook, may I fix this a bit later? I would like to first have this working in an internal testing env.

adapter-kafka/pom.xml
MMirelli: https://github.com/nosqlbench/nosqlbench/pull/1143#issuecomment-1450375869

eolivelli: I didn't mean to add a comment on the PR but to also drop a line like:

```
    <!-- This dependency is needed in order to re-use some utilities from the Pulsar adapter code -->
```

adapter-kafka/pom.xml
MMirelli: good catch, thank you!

adapter-kafka/pom.xml
MMirelli: Oh ok fixed.

adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaConsumer.java
MMirelli: In case there is not such a key among the header keys, a NPE would be thrown, as `value()` is applied on `null`.

eolivelli: so we should handle that case (null) the same way as a blank string

yabinmeng (changes requested):
LGTM overall with some comments. I'm not really a fan of adding dependency among NB modules.

adapter-kafka/pom.xml
MMirelli: @jshook @yabinmeng - I didn't give enough context about this dep. 

I was suggesting to create a new module as a shared util module (generating metrics tracking message duplicates, message loss and re-ordered messages) as this could be easily reused by some module other than `adapter-pulsar` and `adapter-kafka`. For example a module which might need that in the future is `adapter-s4j`.

adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaConsumer.java
MMirelli: Now, I skip the metrics generation in case it's null and warn the user about it.

adapter-kafka/pom.xml
MMirelli: I decoupled moved the reused code in `adapters-api` module in `io.nosqlbench.engine.api.metrics.ReceivedMessageSequenceTracker`. Also see: https://github.com/nosqlbench/nosqlbench/pull/1143/commits/b33389824980c41c7d227a465ffa5875287763d1.

adapter-kafka/src/main/java/io/nosqlbench/adapter/kafka/ops/OpTimeTrackKafkaProducer.java
```diff
@@ -193,6 +206,11 @@ void cycleMsgProcess(long cycle, Object cycleObj) {
         }
 
         ProducerRecord<String, String> message = (ProducerRecord<String, String>) cycleObj;
+        if (seqTracking) {
```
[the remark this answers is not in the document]
MMirelli: I think this is ok for now. We can modify this later on, if needed.

MMirelli:
I also: 
1. tested that I didn't break the pulsar adapter metrics by using https://github.com/nosqlbench/nosqlbench/pull/1143/commits/8eb9498fa98944941d118d507dc732cb71832749 and https://github.com/nosqlbench/nosqlbench/pull/1143/commits/0a2ab1dcca00f1dab3409f4956cb71965e7322c7
2. fixed a change which was breaking e2e error metrics in the pulsar adapter: https://github.com/nosqlbench/nosqlbench/pull/1143/commits/971c5d3c6f6bfec5ebf1d103e6d936b19646c347
3. documented the document level e2e options: https://github.com/nosqlbench/nosqlbench/pull/1143/commits/7d5ecf2f32bd93e58d79d6b5279869ff28fc6662
4. added handy scripts and configuration files to quickly build and test out the adapters: https://github.com/nosqlbench/nosqlbench/pull/1143/commits/8eb9498fa98944941d118d507dc732cb71832749 and https://github.com/nosqlbench/nosqlbench/pull/1143/commits/0a2ab1dcca00f1dab3409f4956cb71965e7322c7

@jshook and @yabinmeng, would it be fine to keep the commits in 4?

MMirelli:
>@MMirelli I'm assuming this is still in the works, since it is still in "changes requested" state.

Unless you have more comments, I think I am done with this.

yabinmeng (commented):
I would recommend to seek Shooky's review since made some changes in the engine code.

MMirelli:
@jshook I would merge this, it's been fairly stable with up to 20K msg/s with transaction enabled.

Issue #948, opened 2023-01-25
Full End-to-End Testing with TestContainers


jshook:
### Discussed in https://github.com/nosqlbench/nosqlbench/discussions/947

<div type='discussions-op-text'>

<sup>Originally posted by **jshook** January 25, 2023</sup>
We have many integrated examples of workloads across a variety of drivers. These are shipped along with the main artifacts.
They have not all been thoroughly tested for each release.

In keeping with our plans to bring these workloads forward into the docs as living examples, we need a way to verify that they all function correctly before we ship a new release.

We can use https://www.testcontainers.org/ to stand up appropriate targets for most or all of our driver types. We can also use the built-in discovery mechanisms like `--list-workloads` to find all examples and run them with a small cycle count.

The goal will be to run all examples at least once before promoting any major release.

</div>

jshook:
Here are a few thoughts on how to tackle this.

The ability to run each of the drivers within the main NB machinery will depend on having the driver and the runtime together. This is provided with nb5. Thus, the reason for the nb5-proof module depending on nb5 as the place to run these test.

However, each driver that is provided should be able to provide its own end-to-end workload. These driver modules will not have the ability execute their workloads during their builds because they are implementing the API portions, not the core, and the core is bundled separately via loose coupling.

So here's a sketch:

- The nb5-proof module will need to use the built-in workload discovery tools (what --list-workloads) is based on, at least, to find all the built-in workloads to iterate over.
- It will need to recognize some signature within each workload for a set of sanity checking defaults. The lack of something like this will likely become a filter for not including the workloads downstream.
- It will need to recognize what (nb5) driver to use, but that should come for free with properly defined _named scenarios_
- Each driver will have to have some corresponding testcontainers image type. This is fine to put in the nb5-proof module, as it's responsibilities include setting up relevant integrated test apparatus before running a workload. Maybe we allow each workload or driver to specify a compatible docker image.

jshook:
Maybe we can look for a suitable docker run command line in each workload and sanitize it. We can extract those parameters as long as they follow strict forms. This might be the best of both worlds for end-users as well as the end-to-end test system.

PR #1145, opened 2023-03-01
Timeout correction for GHA on nosqlbench #948 testcontainers

MikeYaacoubStax:
Increased the timeout for GHA to succeed in running the testcontainers.

Note that GHA is significantly slower than running these tests locally.

Here is the succeeded run on GHA that took nb5-proof testcontainers 5min30sec : https://github.com/nosqlbench/nosqlbench/actions/runs/4304834212


jshook (approved):
LGTM

Issue #1197, opened 2023-04-03
Indentation Errors on Auto-Generated Binding Functions


MikeYaacoubStax:
**Description**
Some auto-generated binding functions have unnessary nested indentations, these indentations are causing some pages to load slowly, and some like 'funcref_collections' to not load at all.

**To Reproduce**
go to: [funcref-collections](https://docs.nosqlbench.io/reference/bindings/funcref-collections/)

**What was Expected**
A well indented page that does not hang on any browser

**Additional context**
Upon investigation: this has to do with some patterns of code used in the automatically generated markdown that indicate an indentation to zola (some combinations of '>' with other characters)

**Screenshots*
![image](https://user-images.githubusercontent.com/117678633/229537318-3ef91a62-370c-4124-8191-900d1836220f.png)

jeffbanks:
Resolved in PR #1198

PR #1198, opened 2023-04-03
PR-NB5-Docs Indentation Fix

MikeYaacoubStax:
This is in response to issue #1197 

Wrapped culprit generated code with code delimiters that way Zola does not perceive any characters as special characters for indentation.

This fix has been validated by running `export-docs` locally and replacing the `funcref_collections` markdown with the newly generated one that includes the code delimiters. The new file solved the issue.


jeffbanks (approved):
LGTM.  Thank you for tackling this one, it's been a nuisance for our docs viewers.

PR #1234, opened 2023-04-27
Metrics Publishing

jeffbanks:
Provides Prometheus exposition formatted metrics to an external endpoint.


jeffbanks:
A huge collective effort on this one.

Thanks for all the collaboration @jshook and @MikeYaacoubStax!  

This change will set us up for even more advanced metrics capabilities in the future.

PR #1251, opened 2023-05-05
Verify expected result with MVEL

kijanowski:
This PR adds a new functionality that allows to verify a feature of Accord transactions - the `SELECT` statement.

Here's a sample Accord transaction:
```
        BEGIN TRANSACTION
        LET acc = (SELECT * FROM <<keyspace:baselines>>.<<table:accounts>> WHERE id = {id});
        SELECT acc.value;
        IF acc.value = 0 THEN
            UPDATE <<keyspace:baselines>>.<<table:accounts>> SET value = 1 WHERE id = {id};
        END IF
        COMMIT TRANSACTION;
```
In this transaction one row is selected into the `acc` field.
Next the SELECT statement determines what will be returned to the client. In this case it is the `value` column.
Then the transaction is executed where the `value` is set to 1 and the transaction is committed.
In NoSQLBench we want to verify, that the `SELECT` statement returned 0 for the `value` column.

This can be achieved with the new `expected-result` workload parameter. The value is a MVEL expression:
```
  params:
    expected-result: "get(0).getInt('acc.value') == 0"
```
or:
```
params:
  expected-result: "get(0).getInt('k') == 0 && get(0).getInt('c') == 0 && get(0).getInt('v') == 0 && get(1).getInt('k') == 0 && get(1).getInt('c') == 1 && get(1).getInt('v') == 1"
```

The new parameter allows to verify the result consisting of a list of rows. The first `get(idx)` retrieves a given row from the result. Then `getInt(id)`, `getLong(id)`, etc can be used to extract a given field by name and type.

There is one open question left - should the logs print the result when it did not meet the expected one.
I tried:
- ReflectionToStringBuilder ( + RecursiveToStringStyle):
```
com.datastax.oss.driver.internal.core.cql.DefaultRow@6a9286f6[data=com.datastax.oss.protocol.internal.util.collection.NullAllowingImmutableList@5aceb007{java.nio.HeapByteBuffer@38917523[bigEndian=true,hb={0,0,0,1},isReadOnly=false,nativeByteOrder=false,offset=0,address=16,capacity=4,limit=4,mark=-1,position=0,segment=<null>]},definitions=com.datastax.oss.driver.internal.core.cql.DefaultColumnDefinitions@19066abd[definitions=com.datastax.oss.driver.shaded.guava.common.collect.SingletonImmutableList@5316e3f9{com.datastax.oss.driver.internal.core.cql.DefaultColumnDefinition@2a8d2cd3[keyspace=com.datastax.oss.driver.api.core.CqlIdentifier@b217b36[internal=accord2],name=com.datastax.oss.driver.api.core.CqlIdentifier@21d8bb6d[internal=acc.value],table=com.datastax.oss.driver.api.core.CqlIdentifier@6270e031[internal=accounts],type=com.datastax.oss.driver.internal.core.type.PrimitiveType@61dcead[protocolCode=9]]},index=com.datastax.oss.driver.internal.core.data.IdentifierIndex@665fd93[byCaseInsensitiveName=com.datastax.oss.driver.shaded.guava.common.collect.LinkedListMultimap@6ed96180[],byCaseSensitiveName=com.datastax.oss.driver.shaded.guava.common.collect.LinkedListMultimap@21909bd[],byId=com.datastax.oss.driver.shaded.guava.common.collect.LinkedListMultimap@5d25e8f8[]]]]
```
- GSON:
```
[
  {
    "definitions": {
      "definitions": [
        {
          "keyspace": {
            "internal": "accord2"
          },
          "table": {
            "internal": "accounts"
          },
          "name": {
            "internal": "acc.value"
          },
          "type": {
            "protocolCode": 9
          }
        }
      ],
      "index": {
        "byId": {},
        "byCaseSensitiveName": {},
        "byCaseInsensitiveName": {}
      }
    },
    "data": [
      {
        "hb": [
          0,
          0,
          0,
          1
        ],
        "offset": 0,
        "isReadOnly": false,
        "bigEndian": true,
        "nativeByteOrder": false,
        "mark": -1,
        "position": 0,
        "limit": 4,
        "capacity": 4,
        "address": 16,
        "segment": null
      }
    ]
  }
]
```

Maybe, instead of pretty-printing the whole response (or just a subset of it), it would be sufficient to evaluate the left side of the MVEL expression that failed the verification? Something like:
`MVEL.executeExpression(MVEL.compileExpression(((BinaryOperation)((ExecutableAccessor) expectedResultExpression).getNode()).getLeft().toString()), result)` - that would return the value that failed the verification. But this approach would assume that every mvel expression is a `BinaryOperation`.

Another option is to simply run `toString()` on the result object. In case of a CQL Row, the result is meaningless, but then instead of returning a list of CQL Rows in `Cqld4CqlOp` we could return a wrapper that knows how to pretty-print itself.

I'm leaving that for discussion, not critical for this PR I think?


kijanowski:
@jeffbanks 
That would be possible, if we wrap the result (an `ArrayList<Row>`) by a `Map`:
```
Map<String, Object> wrpr = new HashMap<>();
wrpr.put("rows", results.get());
return MVEL.executeExpression(expectedResultExpression, wrpr, boolean.class);
```
Instead of `get(0).getInt('accValue')` we now would have `rows[0].getInt('acc.value')` - more readable but at the cost of a new object. The name `rows` would need to be either hardcoded or another parameter on the workload.

jeffbanks:
> Instead of `get(0).getInt('accValue')` we now would have `rows[0].getInt('acc.value')` - more readable but at the cost of a new object. The name `rows` would need to be either hardcoded or another parameter on the workload.

Right, this may not be worth the effort, as we still have to index another array.  No worries, thought it was worth looking at to reduce the verbosity and increase readability.

engine-api/src/main/java/io/nosqlbench/engine/api/activityapi/errorhandling/modular/handlers/ExpectedResultVerificationErrorHandler.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.function.Supplier;
+
+/**
+ * The expected result verification error handler will create, if needed, two metric
+ * objects for error and retry counts.
+ */
+@Service(value = ErrorHandler.class, selector = "verifyexpected")
```
jshook, line 34: I like this capability, but there is probably a better way for us to structure the handler. We can come back to this as a separate question after merge.

engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLIOptions.java
```diff
@@ -340,6 +350,9 @@ private LinkedList<String> parseGlobalOptions(final String[] args) {
                     showStackTraces = true;
                     arglist.removeFirst();
                     break;
+                case NBCLIOptions.ENABLE_DEDICATED_VERIFICATION_LOGGER:
+                    enableDedicatedVerificationLogger();
```
jshook, line 354: There is a bug here. Args must be consumed explicitly on matching.

jshook (changes requested):
Only one small issue to address before merge.

engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLIOptions.java
kijanowski: Adding `arglist.removeFirst();`

jshook (approved):
LGTM! merging

Issue #1260, opened 2023-05-09
Option for setting SSL Validation False Flag


ksmaung:
**What is the ideal scenario for using this feature?**
Performance testing for non-prod data

**Is your feature request related to a problem? Please describe.**
For Cassandra deployment with 40+ clusters with self-signed certificates, it would be impractical, if not impossible, to create trust store certificates for each node for establishing connections to run NoSqlBench.  This option is already provided by the Cassandra Java driver.
Propose to add “sslvalidation=false” to bypass SSL validation to relieve the burden of creating trust store certificate for every node. 

**Would you like to help implementing or testing it?**
We would be happy to implement it if there’s no takers as it is needed sooner.

**Additional context**
Current supported options for SSL
nb help SSL

help ssl
# SSL
 
Supported options:
 
- **ssl** - specifies the type of the SSL implementation.
    Disabled by default, possible values are `jdk`, and `openssl`.
 
- **tlsversion** - specify the TLS version to use for SSL.
 
    Examples:
    - `tlsversion=TLSv1.2` (the default)
 
For `jdk` type, the following options are available:
 
- **truststore** - specify the path to the SSL truststore.
 
    Examples:
    - `truststore=file.truststore`
 
- **tspass** - specify the password for the SSL truststore.
 
    Examples:
    - `tspass=truststore_pass`
 
- **keystore** - specify the path to the SSL keystore.
 
    Examples:
    - `keystore=file.keystore`
 
- **kspass** - specify the password for the SSL keystore.
 
    Examples:
    - `kspass=keystore_pass`
 
- **keyPassword** - specify the password for the key.
 
    Examples:
    - `keyPassword=password`
 
 
For `openssl` type, the following options are available:
 
- **caCertFilePath** - path to the X509 CA certificate file.
 
    Examples:
    - `caCertFilePath=cacert.crt`
 
- **certFilePath** - path to the X509 certificate file.
 
    Examples:
    - `certFilePath=ca.pem`
 
- **keyFilePath** - path to the OpenSSL key file.
 
    Examples:
    - `keyFilePath=file.key`

jeffbanks:
Hi @ksmaung, after looking at this in more detail, can you provide some more information about how you want to bypass the validation in our current set of supported options for `jdk` and `openssl` as they apply to keystore and truststore settings?  

When applying the `sslvalidation=false`, are you expecting that any `truststore` settings would be completely ignored, but the `keystore` settings would still be verified?

**Example**
```
[...]

PR #1289, opened 2023-05-23
Bypass SSL validation option

jeffbanks:
Addresses request #1260.


nb-api/src/main/java/io/nosqlbench/api/engine/util/SSLKsFactory.java
```diff
@@ -81,18 +84,25 @@ public SSLContext getContext(NBConfiguration cfg) {
...
+
+            boolean activeSSLValidation = true;
+            if (cfg.getOptional("sslValidation").isPresent() && cfg.getOptional("sslValidation")
+                    .get().equals("false")) {
+                logger.trace("sslValidation=false, skipping SSL validation.");
+                activeSSLValidation = false;
+            }
 
-            if (sslParam.get().equals("jdk")) {
+            if (activeSSLValidation && sslParam.get().equals("jdk")) {
```
jshook, line 98: It's not clear to me how this is achieving the same effect as suggested by the initial issue. It looks like if you don't enable validation that we completely disregard trust store init. Since changes here are difficult to test in myriad configurations, I want to be extra careful with this are of the code.

jshook (commented):
I think we need more detail from the original user about exactly what validation to disable, and whether leaving the trust store unconfigured has the same effect. This will require some on- branch testing with a few configurations before we can merge it.

jshook (approved):
This version LGTM! It connects the dots nicely to the original request.

PR #1291, opened 2023-05-24
Filter script parameters from activity context

smccarthy788:
Found while using a variant of `stepup.js`. When using `params.withDefaults` in a scenario script, all parameters given on the command line are passed into the activity context. This can cause conflicts with the parameter checking, resulting in errors similar to:
```
javax.script.ScriptException: io.nosqlbench.api.errors.BasicError: Unknown config parameter 'rate_step' in config model while configuring StandardActivity, possible parameter names are
```

This change introduces the `filterScriptParams` function to scenario scripts which use `params.withDefaults`. I've included `SCRIPT_PARAMS` for each script, but have only read the scripts to judge what those parameters are.

`filterScriptParams` was tested against the aformentioned variant of `stepup.js` and shown to resolve the error seen.

### DIFF
--- a/engine-cli/src/main/resources/scripts/auto/fm.js
+++ b/engine-cli/src/main/resources/scripts/auto/fm.js
@@ -2,6 +2,24 @@
 function printa(arg) { java.lang.System.out.print(arg); }
 //function println(arg) { java.lang.System.out.println(arg); }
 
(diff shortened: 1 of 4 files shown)


jshook (approved):
LGTM. I would like us to handle this in the runtime more gracefully, so https://github.com/nosqlbench/nosqlbench/issues/1297

PR #1296, opened 2023-05-25
Add initial vector variate generation functions

### LINKED ISSUE #1295: Add vector generation bindings
We need support functions to generate deterministic vector data. This should eventually include unit vectors, dimensions, and distributions  over individual or multiple dimensions or variance from a reference vector.

jshook:
closes #1295

### DIFF
--- a/virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/NormalizeVector.java
+++ b/virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/NormalizeVector.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2023 nosqlbench
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.nosqlbench.virtdata.library.basics.shared.from_long.to_vector;
+
+import io.nosqlbench.virtdata.api.annotations.Categories;
+import io.nosqlbench.virtdata.api.annotations.Category;
+import io.nosqlbench.virtdata.api.annotations.ThreadSafeMapper;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * Normalize a vector.
+ */
+@ThreadSafeMapper
+@Categories(Category.experimental)
+public class NormalizeVector implements Function<List<Double>,List<Double>> {
+    @Override
+    public List<Double> apply(List<Double> doubles) {
+        ArrayList<Double> unit = new ArrayList<>(doubles.size());
+        double accumulator = 0.0d;
+        for (Double scalar : doubles) {
+            accumulator+=scalar*scalar;
+        }
+        double vectorLen = Math.sqrt(accumulator);
+        for (double scalarComponent : doubles) {
+            unit.add(scalarComponent/vectorLen);
+        }
+        return unit;
+    }
+}
--- a/virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/NormalizeVector.java
+++ b/virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/NormalizeVector.java
@@ -37,9 +37,9 @@ public class NormalizeVector implements Function<List<Double>,List<Double>> {
         for (Double scalar : doubles) {
             accumulator+=scalar*scalar;
         }
-        double vectorLen = Math.sqrt(accumulator);
+        double scalarLen = Math.sqrt(accumulator);
         for (double scalarComponent : doubles) {
-            unit.add(scalarComponent/vectorLen);
+            unit.add(scalarComponent/scalarLen);
         }
         return unit;
     }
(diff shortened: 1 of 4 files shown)


virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/NormalizeVector.java
```diff
@@ -0,0 +1,46 @@
...
+@Categories(Category.experimental)
+public class NormalizeVector implements Function<List<Double>,List<Double>> {
+    @Override
+    public List<Double> apply(List<Double> doubles) {
+        ArrayList<Double> unit = new ArrayList<>(doubles.size());
+        double accumulator = 0.0d;
+        for (Double scalar : doubles) {
+            accumulator+=scalar*scalar;
+        }
+        double vectorLen = Math.sqrt(accumulator);
```
ShaunakDas88, line 40: nit: name this vectorNorm? vector length can be potentially misinterpreted as the number of components

jeffbanks (approved):
lgtm; nice addition

ShaunakDas88 (approved):

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/UniformVectorSizedStepped.java
```diff
@@ -0,0 +1,76 @@
...
+        } else {
+            dim=dims.length/2;
+        }
+        if ((dims.length%2)!=0) {
+            throw new RuntimeException("Unable to set uniform range as [min,max] for pairs when count is odd. You must provide complete [min, max] value pairs as a,b,c,d,...");
+        }
+        this.funcs = new LongToDoubleFunction[dim];
+        for (int i = 0; i < dim; i++) {
+            if (i<dims.length/2) {
+                funcs[i]=new HashedDoubleRange(dims[i<<1].doubleValue(),dims[i<<1].doubleValue()+1);
```
ShaunakDas88, line 61: Question: the bit shift to the right doubles index `i`. Should the second argument in HashedDoubleRange be:
```
dims[(i<<1) + 1].doubleValue()
```
? Perhaps I'm misunderstanding something.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/UniformVectorSizedStepped.java
```diff
@@ -0,0 +1,76 @@
...
+    private final int dim;
+    private final LongToDoubleFunction[] funcs;
+
+    @Example({"UniformVectorSizedStepped(3)","create a 3-component vector from unit interval variates"})
+    @Example({"UniformVectorSizedStepped(1.0d,100.0d,5.0d,6.0d)","create a 2-component vector from the specified uniform ranges [1.0d,100.0d] and [5.0d,6.0d]"})
+    @Example({"UniformVectorSizedStepped(2,3.0d,6.0d)","create a 2-component vector from ranges [3.0d,6.0d] and [0.0d,1.0d]"})
+    public UniformVectorSizedStepped(Number... dims) {
+        if (dims.length>=1 && (dims.length)%2==1 && dims[0] instanceof Integer) {
+            this.dim = dims[0].intValue();
+            dims = Arrays.copyOfRange(dims,1,dims.length);
```
ShaunakDas88, line 51: minor performance nit: Rather than making a copy from index 1 on of the array, perhaps a `startOffset` pointer to index 1 here (and index 0 in the else). Then the true length we care about is `dims.length - startOffset`, and we can just iterate through the original dims appropriately to index arithmetic?

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/UniformVectorSizedStepped.java
jshook: It looks to me like you found a bug, and either the passing test is a mathematical coincidence, or I've missed coverage here.

jshook: Ah the test was not specific enough. I made it quantitative. The previous version was asserting only a subset of the expected range.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/UniformVectorSizedStepped.java
jshook: I would generally agree, except that in this case the constructor logic is only used during initialzation. Once a binding function is resolved and installed, it is static, and only what happens in the apply method will be perf sensitive. Not opposed to the change, though.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/NormalizeVector.java
jshook: done

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/UniformVectorSizedStepped.java
jeffbanks: 👍🏼 Good catch @ShaunakDas88

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_vector/UniformVectorSizedStepped.java
jeffbanks: @ShaunakDas88 I had a similar question in the past, this is the preferred way for init and apply in the bindings for performance reasons.

ShaunakDas88 (approved):

PR #1306, opened 2023-05-30
Pc driver adapter

MarkWolters:
Fixed bug in pinecone adapter where config model for adapter space was not exposed through the adapter class

### DIFF
--- a/adapter-pinecone/src/main/resources/activities/pinecone-query.yaml
+++ b/adapter-pinecone/src/main/resources/activities/pinecone-query.yaml
@@ -1,0 +1,14 @@
+scenarios:
+  default:
+    query: run driver=pinecone tags==blocks:query cycles=10
+
+blocks:
+  query:
+    ops:
+      op1:
+        query: "test-index"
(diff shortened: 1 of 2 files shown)


adapter-pinecone/src/main/resources/activities/pinecone-query.yaml
```diff
@@ -0,0 +1,14 @@
+scenarios:
+  default:
+    query: run driver=pinecone tags==blocks:query cycles=10
+
+blocks:
+  query:
+    ops:
+      op1:
+        query: "test-index"
+        vector: "0.8602578079921012,0.12103044768221516,0.7737329191858439,0.4521093269320254,0.29351661477669416,0.4261807015226558,0.14131665592103335,0.882370813029422,0.4412833140430886,0.9916525700115515"
```
jshook, line 10: Not a blocker here, but we need to allow this to be structurally a list of numbers too.
As a follow-on we can test an example value by applying the function, and then augment the composed function with the right kind of casting and conversion to match.

jshook (approved):
Just notes for a future change

jeffbanks (approved):
lgtm

PR #1308, opened 2023-05-31
Initial complete draft of NB5 S4R (no REAMD.md yet)

yabinmeng:


adapter-s4r/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+        into an AMQP-0.91 (e.g. RabbitMQ) or an AMQP-0.91-compatible (e.g. Pulsar with S4R) system.
+    </description>
+
+    <properties>
+        <amqp.version>5.17.0</amqp.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.nosqlbench</groupId>
```
jshook, line 42: This dependency is not needed here. If it is, there is a refactoring that should happen upstream.

PR #1311, opened 2023-06-02
Pc driver adapter

MarkWolters:
Added capability to define vector input as csv separated strings or as lists of type string, float, or double. 
Cleaned up code and moved some common functionality to base PineconeOpDispenser class.
Allow metadata and filter types of string, numeric or list of string.


adapter-pinecone/src/main/java/io/nosqlbench/adapter/pinecone/opdispensers/PineconeOpDispenser.java
```diff
@@ -73,6 +76,60 @@ protected ListValue generateListValue(List comparator) {
...
+        List<Float> floatValues;
+        if (rawVectorValues instanceof String) {
+            floatValues = new ArrayList<>();
+            String[] rawValues = (((String) rawVectorValues).split(","));
+            for (String val : rawValues) {
+                floatValues.add(Float.valueOf(val));
+            }
+        } else if (rawVectorValues instanceof List) {
+            floatValues = switch (((List<?>) rawVectorValues).get(0).getClass().getSimpleName()) {
+                case "Float" -> (List<Float>) rawVectorValues;
```
jshook, line 108: Ok for now, but we should pre-qualify the type from a test value and make this not do type reflection on every call to the op dispenser. Once you know a given op template field produces a certain base value type, you can assume it continue to do the same. In other words, The instanceof checks should only be called in the op mapper or in the op dispenser constructor.

jshook (approved):
I think this is ready to merge. Some notes added for improvements later.

jeffbanks (approved):
lgtm, I don't see anything that stands out for this initial version.  Extremely nice work @MarkWolters!

PR #1316, opened 2023-06-06
Vector search java-driver integration

jeffbanks:
Includes changes required for integration with vector search inclusions in DataStax oss [java-driver](https://github.com/datastax/java-driver).

Specifically, changes in [this PR](https://github.com/datastax/java-driver/pull/1639).


adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/opdispensers/CQLD4PreparedStmtDiagnostics.java
```diff
@@ -50,84 +52,82 @@
...
-                    Optional<Map.Entry> first = entries.stream().findFirst();
-                    if (first.isPresent()) {
-                        yield bound.setMap(colname, map, first.get().getKey().getClass(), first.get().getValue().getClass());
-                    } else {
-                        yield bound.setMap(colname, map, Object.class, Object.class);
-                    }
+        return switch (coltype.getProtocolCode()) {
+            case CUSTOM -> {
+                if (coltype instanceof CqlVectorType) {
+                    yield bound.setCqlVector(colname, (CqlVector<?>) colval);
```
jshook, line 63: Nice add. It would be nice to come back to this for any other packaged Custom types.

adapter-cqld4/src/main/java/io/nosqlbench/datamappers/functions/geometry/NormalizeCqlVector.java
```diff
@@ -0,0 +1,69 @@
...
+        CqlVector.Builder builder = CqlVector.builder();
+        Iterable values = cqlVector.getValues();
+        List<Object> list = new ArrayList<>();
+        values.forEach(list::add);
+
+        if (list.isEmpty()) {
+
+            builder.add(List.of());
+        } else if (list.get(0) instanceof Float) {
+            List<Float> floats = new ArrayList<>();
```
jshook, line 52: A follow-up change to the driver is expected to simplify this and make it more efficient. We should include it in any follow-ups as well.

jshook (approved):
I re-reviewed everything and didn't see anything which was a good reason not to merge this.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/conversions/from_long/ToDouble.java
```diff
@@ -0,0 +1,59 @@
...
+@Categories(Category.conversion)
+@ThreadSafeMapper
+public class ToDouble implements LongToDoubleFunction {
+
+    private final LongToDoubleFunction func;
+
+    ToDouble(Object func) {
+        if (func instanceof Double aDouble) {
+            double scale = aDouble.doubleValue();
+            this.func = l -> (double) (l % scale);
```
ShaunakDas88, line 46: Reading the comments:
```
 If a double is provided, then the long input is scaled to be between 0.0f and the value provided, as a fraction of the highest possible long.
 ```
Mod will get us between 0 and `scale`. But can you explain the `scaled...as a fraction of the highest possible long`part?

File: virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/conversions/from_long/ToFloat.java [hunk 4]
```diff
@@ -19,23 +19,41 @@
...
-        this.scale = scale;
-    }
-    public ToFloat() {
-        this.scale = Long.MAX_VALUE;
+    private final LongFunction<Float> func;
+
+    ToFloat(Object func) {
+        if (func instanceof Float afloat) {
+            float scale = afloat.floatValue();
+            this.func = l -> (float) (l % scale);
```
ShaunakDas88, line 46: similar question as `ToDouble`.

ShaunakDas88 (commented):
There were quite a few files, so hopefully I didn't miss anything major. Just a few comments.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_float/HashedFloatRange.java
```diff
@@ -0,0 +1,73 @@
...
+public class HashedFloatRange implements LongFunction<Float> {
+
+    private final float min;
+    private final float max;
+    private final float interval;
+    private final static float MAX_FLOAT_VIA_LONG = (float) Long.MAX_VALUE;
+    private final Hash hash = new Hash();
+
+    public HashedFloatRange(float min, float max) {
+        this.min = min;
```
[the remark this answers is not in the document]
jshook: This is missing, and should be added. The Hash takes abs(result) before using it here, but the ranging logic is not defensive enough, so you could possibly use this with invalid values.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_float/HashedFloatRange.java
```diff
@@ -0,0 +1,73 @@
...
+        this.max = max;
+        this.interval = max - min;
+    }
+
+    @Override
+    public Float apply(long value) {
+        long hashed = hash.applyAsLong(value);
+        float unitScale = ((float) hashed) / MAX_FLOAT_VIA_LONG;
+        float valueScaled =interval*unitScale + min;
```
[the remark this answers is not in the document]
jshook: Yes, line 69 creates a unit-interval variate, with some positive long >= 0 and <= (float) Long.MAX_VALUE, and then maps that unit-inverval to a scaled interval stacked on the minimum.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/from_long/to_float/HashedRangedToNonuniformFloat.java
```diff
@@ -0,0 +1,73 @@
...
+    private final long min;
+    private final long max;
+    private final float length;
+    private final Hash hash;
+
+    public HashedRangedToNonuniformFloat(long min, long max) {
+        this.hash = new Hash();
+        if (max<=min) {
+            throw new RuntimeException("max must be >= min");
+        }
```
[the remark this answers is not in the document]
jshook: agreed, these are needed everywhere we have *Range or *Interval functions.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/conversions/from_long/ToDouble.java
jshook: The docs should def be improve here.

What this means is, that since the input is always a long, that is the "typed unit inteval" we have to work with, i.e. <long variate> / <maximum long value> as a fractional magnitude. Based on that, if you want to produce a double value, it will be by mapping (0.0d/(double) Long.MAX_VALUE) a a unit interval to a double.

This was an attempt to provide multiple options for how this ToDouble is used, but I think it is overly complicated, given that the variations do actually have different behavior. We should separate these.

virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/conversions/from_long/ToFloat.java
jshook: Same problem, same remedy

jshook (approved):
if this passes another review, can we merge it?

ShaunakDas88 (approved):

PR #1353, opened 2023-06-24
Update NBS4J version to 4.0.1 

yabinmeng:
1) Since NBS4J  version 4.0.1 changes to use OSS Pulsar client version 3.0.0 as the underlying dependency, this PR also upgrades the NB Pulsar client version dependency to 3.0.0 (mixed Pulsar client versions are known to cause issues when running NB S4J tests)

2) Also add a convenient script for end-to-end sanity NB S4J workload tests. This is helpful to detect errors like (runtime) version conflict.


eolivelli (approved):
This is great.

I wonder if we could add an integration/unit test that starts Pulsar with docker (Luna Streaming + s4k) and run the sanity check script in CI.

In the S4K report there are examples (DockerTest) about how to start various versions of Pulsar with s4k on docker

msmygit (approved):

adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/util/S4JClientConf.java
```diff
@@ -156,12 +158,9 @@ else if (StringUtils.startsWith(confKey, JMS_CONF_PREFIX)) {
                     }
                 }
             }
-        } catch (IOException ioe) {
-            logger.error("Can't read the specified config properties file: " + fileName);
-            ioe.printStackTrace();
-        } catch (ConfigurationException cex) {
-            logger.error("Error loading configuration items from the specified config properties file: " + fileName + ":" + cex.getMessage());
-            cex.printStackTrace();
+        } catch (IOException | ConfigurationException ex) {
```
jshook, line 161: Is there an exception type leak here? In other words if an exception besides these two were to happen, is it handled as expected?

adapter-s4j/src/main/resources/sanity-validation/e2e-nbs4j-sanity.sh
```diff
@@ -0,0 +1,147 @@
+#! /usr/local/bin/bash
```
jshook, line 1: I'm not sure we should be adding shell scripts as assets here. Ideally, everything you need to run a complex workload can be put into named scenarios. What is the case for this one being unique enough that a different approach is needed?

adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/util/S4JClientConf.java
yabinmeng: I believe the code snippet above only throws out these 2 types of exceptions.

adapter-s4j/src/main/resources/sanity-validation/utilities.sh
```diff
@@ -0,0 +1,113 @@
+#! /usr/local/bin/bash
```
jshook, line 1: same concern as with the other shell code. Essentially, this is an opaque type which nosqlbench doesn't understand and can't verify, validate, use directly, or marshal. It is surely a security vector.

adapter-s4r/src/main/resources/sanity-validation/e2e-nbs4r-sanity.sh
```diff
@@ -0,0 +1,137 @@
+#! /usr/local/bin/bash
```
jshook, line 1: Let's work together to put these in native nb scenario script form or CLI script form.

jshook (commented):
I'd like to remove the (increasingly more) elaborate shell code here. There is nothing here that shouldn't be in a scenario script, or a set of named scenarios. If I'm mistaken, let's get together soon and discuss.

adapter-s4j/src/main/resources/sanity-validation/utilities.sh
yabinmeng: Deleted the bash scripts (as per our discussion)

adapter-s4r/src/main/resources/sanity-validation/e2e-nbs4r-sanity.sh
yabinmeng: Deleted the bash scripts (as per our discussion)

jshook (approved):
Looks good to me

PR #1365, opened 2023-06-27
Add named scenarios for S4J and S4R. Update the README doc accordingly.

yabinmeng:


adapter-s4j/src/main/resources/scenarios/nbs4j_msg_proc_named.yaml
```diff
@@ -0,0 +1,28 @@
+scenarios:
+  msg_send:
+    - run driver=s4j cycles=100 threads=4 num_conn=2 num_session=2 session_mode=client_ack config=../conf/s4j_config.properties service_url=pulsar://localhost:6650 web_url=http://localhost:8080
```
jshook, line 3: It would be preferable to use named steps. To do this, simply replace the list with a map, where the keys are the step names.

jshook (approved):
I would strongly recommend modifying the named scenarios to use named steps. Other than that, LGTM

adapter-s4j/src/main/resources/s4j.md
```diff
@@ -39,7 +51,7 @@ Other NB engine parameters are straight forward:
 
 # 3. NB S4J Driver Configuration Parameter File
 
-The S4J API has a list of configuration options that can be found here: https://docs.datastax.com/en/fast-pulsar-jms/docs/1.1/pulsar-jms-reference.html#_configuration_options.
+The S4J API has a list of configuration options that can be found here: https://docs.datastax.com/en/streaming/starlight-for-jms/3.2/reference/pulsar-jms-reference.html.
```
msmygit, line 54: #nitpick I'd change this to `latest` so it stays future-proofed.
```suggestion
The S4J API has a list of configuration options that can be found here: https://docs.datastax.com/en/streaming/starlight-for-jms/latest/reference/pulsar-jms-reference.html.
```

msmygit (approved):
+1 to the named scenarios.

adapter-s4j/src/main/resources/s4j.md
yabinmeng: Thanks @msmygit. Fixed

adapter-s4j/src/main/resources/scenarios/nbs4j_msg_proc_named.yaml
yabinmeng: Thanks. Changed to named steps.

PR #1375, opened 2023-06-30
Use further compression to logs & code coverage reports

msmygit:
I'm hoping that this further compresses and reduces the space consumed.


jeffbanks (approved):
lgtm.  Very compressive!  😄

PR #1388, opened 2023-07-06
Snyk remediation 07.06.2023

jeffbanks:
A collection of version adjustments to address a number of `high` and `medium` CVEs reported by Snyk.


jshook (changes requested):
For the version changes, those dependencies which are shared between multiple modules should not specify version, and instead version should be specified in full coordinate form in the dependencyManagement section of the mvn-default/pom.xml file.

jeffbanks:
@jshook migration to `mvn-defaults` for the impacted `pom` files.

engine-api/pom.xml
```diff
@@ -92,20 +92,31 @@
...
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk15on</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
```
jshook, line 113: Where did this dependency come from? Let's get rid of guava here!

engine-docker/pom.xml
```diff
@@ -35,27 +35,38 @@
...
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk15on</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
```
jshook, line 60: Same as above. We've worked to avoid guava dependencies due to painful integration problems in the past. If it's back in for some reason, let's see what we can do to limit the impact/exposure.

jeffbanks: @jshook looking at the dependencies, it looks like `
com.github.docker-java:docker-java-core@3.3.1` brings in this vulnerability for `guava` in nb.  

It's considered a low `cve`, so if we want to remove the direct dependency upgrade for `guava` we can keep the `low` marker on the security dashboard.  

Otherwise, the changes for the referenced `guava` version in the PR will remove that low vulnerability identified as: CWE-379: Creation of Temporary File in Directory with Insecure Permissions.

engine-api/pom.xml
jeffbanks: See reply in the comment below.

engine-docker/pom.xml
jeffbanks: 😄 @jshook Looks like another bump of that `docker-java-core` just came in today [here](https://github.com/nosqlbench/nosqlbench/pull/1391). 

Let me refactor those dependencies once again...

jeffbanks: Updated latest with the recent upgrade for `docker-java-core` which has reported to handle much of these specific versions in the original push.

jshook (approved):
LGTM

PR #1392, opened 2023-07-07
SignalPID() binding addition

jeffbanks:
Binding to provided NB process identifier while used as an agent in signal mode.


MikeYaacoubStax (approved):
LGTM! Let's use it in signal!

PR #1447, opened 2023-08-16
Hdf5

MarkWolters:
Adding repositories for hdf-loader and virtdata-lib-hdf5 for, respectively, loading hdf5 files to Astra/file/stdout and nb binding to use hdf5 file as basis for adding records to DSE/Astra


jshook (approved):
Woowoo!

PR #1468, opened 2023-08-22
some vector function updates

jshook:


adapter-cqld4/src/main/java/io/nosqlbench/engine/extensions/vectormath/VectorMath.java
MarkWolters, line 1: looks like these changes are duplicated across PRs

MarkWolters (approved):
changes included in PR1470 also

PR #1490, opened 2023-08-29
recall compatibility

phact:


adapter-http/pom.xml
```diff
@@ -41,6 +41,12 @@
             <version>${revision}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
jshook, line 44: Can we please do this with Gson instead? jackson keeps creating security and compatibility issues for us across modules. Gson is already in the dependency tree and used for many other SERDES needs.

jshook (changes requested):
Please use Gson and banish jackson from this module.

adapter-http/pom.xml
```diff
@@ -41,6 +41,11 @@
...
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.10.1</version>
```
jshook, line 47: minor detail for after, this version should be left to mvn-default management section.

PR #1491, opened 2023-08-29
Pc vectormath

MarkWolters:
Added vector math classes specific to pinecone. Used new package to avoid creating inter-module dependencies.


jshook (approved):
Good for now. We'll do a follow-up to organize these helper libraries in a better way.

PR #1493, opened 2023-08-30
PromPushReporter Configure to Use SSL and BearerToken

### LINKED ISSUE #1482: Modify prom push logic to include SSL and bearer token auth

dave2wave:
Fixes #1482 

This adds a `--prompush` optional configuration. The processing allows configuration of a Bearer token in the file `~/.nosqlbench/prompush/prompush_apikey`

### DIFF
--- a/nb-api/src/main/java/io/nosqlbench/api/engine/metrics/reporters/PromPushReporter.java
+++ b/nb-api/src/main/java/io/nosqlbench/api/engine/metrics/reporters/PromPushReporter.java
@@ -37,6 +37,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


nb-api/src/main/java/io/nosqlbench/api/engine/metrics/reporters/PromPushReporter.java
```diff
@@ -28,32 +31,82 @@
...
     ) {
         super(registry, name, filter, rateUnit, durationUnit);
         uri = URI.create(targetUriSpec);
+        needsAuth = false;
+        ConfigLoader loader = new ConfigLoader();
+        List<Map> configs = loader.load(config, Map.class);
+        NBConfigModel cm = this.getConfigModel();
+        if (configs != null) {
+            logger.info("PromPushReporter process configuration: %s", config);
+            for (Map cmap : configs) {
```
jshook, line 71: This form would represent a number of prom push reporters, so as a revision, this logic should be hoisted out as a wrapper for loading zero or more prom push reporters.

jshook (approved):
Approved with one request to follow

nb-api/src/main/java/io/nosqlbench/api/engine/metrics/reporters/PromPushReporter.java
dave2wave: I think that there is an opportunity to make all of the metric reporters more configurable essentially as a flattened json. This would work more like the annotators where I grabbed that pattern.

PR #1408, opened 2023-07-19
Fix for paging

jeffbanks:
A collection of changes related to the handling of paging logic in `Cqld4CqlOp`.

PR #1497, opened 2023-08-30
Jshook/update paging fix

### LINKED ISSUE #1231: Paging seems broken in nosqlbench5
**Describe the bug**

Paging doesn't seem to work with cql and nosqlbench5, regardless of how high the maxpages
value is, I keep running into the same error.

**To Reproduce**

Spin up a cluster with the following benchmark yaml, trigger the schema, 
rampup and eventually main-read.

```
description: |
  TODO

bindings:
  id: Shuffle(0,10); Template("{}-{}", Hash(), Hash());
  date: Shuffle(0,31); ToString();
  timestamp: AddHashRange(0L,2592000L); Mul(1000); StartingEpochMillis("2023-04-01 12:00:00"); ToJavaInstant();
  value: Shuffle(0,50); Template("value-{}-{}-{}", Hash(), Hash(), Hash());

blocks:
  schema:
    params:
      prepared: false
    ops:
      create-keyspace: |
        CREATE KEYSPACE IF NOT EXISTS TEMPLATE(keyspace, baselines)
        WITH replication = {'class': 'NetworkTopologyStrategy', 'ue1': 'TEMPLATE(rf,3)'}
        AND durable_writes = true;
      create-table : |
        CREATE TABLE IF NOT EXISTS TEMPLATE(keyspace,baselines).TEMPLATE(table,test) (
          id text,
          date text,
          timestamp timestamp,
          value text,
          PRIMARY KEY((id), date, timestamp)
        ) WITH CLUSTERING ORDER BY (date DESC, timestamp DESC)
        AND compaction = { 'class':'LeveledCompactionStrategy' }
        AND compression = { 'class': 'LZ4Compressor', 'chunk_length_in_kb': 4 };
  rampup:
    params:
      cl: TEMPLATE(write_cl,LOCAL_ONE)
      prepared: true
      page_size: 100
      maxpages: 100
    ops: |
      INSERT INTO TEMPLATE(keyspace,baselines).TEMPLATE(table,test) (id, date, timestamp, value)
      VALUES ({id}, {date}, {timestamp}, {value})
  main-read:
    params:
      cl: TEMPLATE(read_cl, LOCAL_ONE)
      prepared: true
    statements:
      select:
        op: |
          SELECT * FROM baselines.test
          WHERE id={id}
          ORDER BY date DESC, timestamp DESC
```

Schema: `./nb5 run workload=bench-small.yaml host={HOST} driver=cql localdc=ue1 tags=block:schema`

Rampup: `./nb5 run workload=bench-small.yaml host={HOST} driver=cql localdc=ue1 cycles=10000 striderate=100 stride=100 tags=block:rampup threads=10`

Read: `./nb5 run workload=bench-small.yaml host={HOST} driver=cql localdc=ue1 cycles=1 striderate=100 stride=100 tags=block:main-read threads=180`

This runs into the following error, regardless of how high we bump maxpages:

```
Error in core motor loop:java.lang.RuntimeException: io.nosqlbench.adapter.cqld4.exceptions.UnexpectedPagingException: Additional paging would be required to read the results from this query fully, but the user has not explicitly indicated that paging was expected. fetched/allowed: 102/100 fetchSize(100): SELECT * FROM baselines.test
```

**What was Expected**

The maximum page rows per partition should be ~1000, so we should be able to page through this
with the given maxpages and page_size.

**Additional context**
OS: Windows, Linux (distribution), macOS ...
environment: k8s, docker, ...
version info (`./nb --version` or `java -jar nb.jar --version`)

**Screenshots, if applicable**

### LINKED ISSUE #1465: Add a custom metrics gauge that tracks recall stats accurately over time
We're currently leveraging HDR histograms instruments with a large multiplier and a standard divisor. This is a bit of a hack. The recall metrics during testing should be provided at the client level as such, so that testing which sends data to a local metrics CSV directory read as what they are, instead of require further downstream derivation.

This should be implemented as additional capabilities, not to replace any other metrics tooling.

This should also be exposed as a scripting plugin. Maybe this is added to the existing scripting metrics plugin as a dynamic defined guage, using other metrics, but in a way that it always acts like a built-in facility. This means that it should be easy to define a floating-point statistics accumulator which doesn't lose data due to histogram reservoir behavior, and which doesn't require encoding in an unfriendly type.
It's probably easiest to implement this using the Java library built-in java.util.DoubleSummaryStatistics.

The extant plugin which should probably be the basis for this lives in
`engine-extensions/src/main/java/io/nosqlbench/engine/extensions/scriptingmetrics`

The metrics instrument doesn't need to implement any of the standard metrics APIs, since it can be it's own POJO/service object. However, it needs to provide a standard Gauge implementation so that it can be seen by the metric registry as a standard metric object, and thus fed downstream where other metrics data flows.

As long as this is implemented as a standard Gauge, then all of the surrounding metrics capabilities should work with it as-is.

### LINKED ISSUE #1494: open metrics expo format is using summaries instead of histograms (for histograms)

### LINKED ISSUE #1495: annotations do not include span:interval|instant for filtering

jshook:
Folded in changes from Jeff's early work on fixing paging and fixed a few other things which were affecting sane dashboarding with the CQL driver.

fixes #1465
fixes #1494
fixes #1495
fixes #1408 
fixes #1231

### DIFF
--- a/adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/optypes/Cqld4CqlOp.java
+++ b/adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/optypes/Cqld4CqlOp.java
@@ -120,7 +120,7 @@ public abstract class Cqld4CqlOp implements CycleOp<List<Row>>, VariableCapture,
         });
 
         try {
-            return rowsStage.toCompletableFuture().get(3, TimeUnit.SECONDS);
+            return rowsStage.toCompletableFuture().get(300, TimeUnit.SECONDS);
         } catch (Exception e) {
             if (e instanceof RuntimeException re) throw re;
             throw new RuntimeException(e);
--- a/adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/optypes/Cqld4CqlOp.java
+++ b/adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/optypes/Cqld4CqlOp.java
@@ -17,21 +17,28 @@
 package io.nosqlbench.adapter.cqld4.optypes;
 
 import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
-import com.datastax.oss.driver.api.core.cql.ResultSet;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.datastax.oss.driver.api.core.cql.Statement;
-import io.nosqlbench.adapter.cqld4.*;
+import io.nosqlbench.adapter.cqld4.Cqld4CqlReboundStatement;
+import io.nosqlbench.adapter.cqld4.LWTRebinder;
+import io.nosqlbench.adapter.cqld4.RSProcessors;
 import io.nosqlbench.adapter.cqld4.exceptions.ChangeUnappliedCycleException;
 import io.nosqlbench.adapter.cqld4.exceptions.ExceededRetryReplaceException;
-import io.nosqlbench.adapter.cqld4.exceptions.UndefinedResultSetException;
 import io.nosqlbench.adapter.cqld4.exceptions.UnexpectedPagingException;
+import io.nosqlbench.adapter.cqld4.instruments.CqlOpMetrics;
 import io.nosqlbench.adapters.api.activityimpl.uniform.flowtypes.*;
+import org.apache.commons.lang3.NotImplementedException;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.util.ArrayList;
-import java.util.Iterator;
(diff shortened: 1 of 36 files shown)


adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/optypes/Cqld4CqlOp.java
```diff
@@ -45,100 +52,102 @@
...
             }
-            totalRows += pageRows;
+            return collect(rs, completeRowSet, cycle);
+        }).exceptionally(throwable -> {
+            if (throwable instanceof RuntimeException tre) throw tre;
+            throw new RuntimeException(throwable);
+        });
+
+        try {
+            return rowsStage.toCompletableFuture().get(3, TimeUnit.SECONDS);
```
jshook, line 123: This should be relaxed to a larger value and/or made configurable.

jshook (commented):
Did a basic review interactively, some minor notes for after

PR #1503, opened 2023-08-31
Jshook/filecache

### LINKED ISSUE #1502: Make large external files automatically cacheable locally

jshook:
fixes #1502 

This allows us to reference urls directly and have them cached locally.
Caveat, this needs to be made safer for concurrent or episodic access patterns, for example when a file is being accessed in mmap mode and another async activity accesses it after the expiry.

        # 10 seconds allows local behavior as before, but re-fetches each start
        # this needs to be improved and made clear for users to avoid Heisenbugs
        --filecache-expiry 10
        # match all patterns by default
        --filecache-pattern ".*"


jshook:
This version is superseded by a better version from @MarkWolters. Removing the PR for now, and the branch shortly.

PR #1505, opened 2023-08-31
modularize methods for verifiers

### LINKED ISSUE #1466: Organize the scripting plugins by adapter, and inject them lazily
With the current scripting plugins, all potential objects are created for every scenario and injected into the scripting environment as a global variable. This is inefficient and problematic. Instead, a service object with the name of `_services` should be used, and all of its members should be resolved lazily using the same scripting plugin mechanisms that already exist.

This should apply equally to the scenario scripting layer as it does to the verifier scripting layer.

We don't want to take out the original service object implementations or integrated testing examples until 
we have the dedicated naming and lazy loader in place.
Even then, the existing scripting service objects should only be deprecated for awhile. They can be removed on a new minor version bump, likely with the 5.21 series, which is due to start after the drop of the next Java LTS 21.

jshook:
fixes #1466


adapter-pinecone/src/main/java/io/nosqlbench/engine/extensions/scriptingutils/PineconeScriptingUtils.java
MarkWolters, line 1: I would like to also add a method such as "metadataFieldToIntArray" as the id field won't necessarily always be an int, that's dependent on the way we are testing currently. Might be useful to be able to use the metadata for the same purpose. Could be in a different pull.

adapter-cqld4/src/main/java/io/nosqlbench/engine/extensions/vectormath/CqlUtilsPluginInfo.java
```diff
@@ -18,19 +18,19 @@
...
 import com.codahale.metrics.MetricRegistry;
 import io.nosqlbench.api.config.LabeledScenarioContext;
-import io.nosqlbench.api.extensions.ScriptingPluginInfo;
+import io.nosqlbench.api.extensions.ScriptingExtensionPluginInfo;
 import io.nosqlbench.nb.annotations.Service;
 import org.apache.logging.log4j.Logger;
 
-@Service(value = ScriptingPluginInfo.class,selector = "vectormath")
-public class VectorMathPluginInfo implements ScriptingPluginInfo<VectorMath> {
+@Service(value = ScriptingExtensionPluginInfo.class,selector = "cql_utils")
```
MarkWolters, line 25: Is the idea truly that this will in the future serve as a plugin for any cql utils or will it continue to be specific to vector stuff? Small quibble but it's named like the former but organized (package wise) like the latter.

MarkWolters (approved):
Looks good. I think there's more functionality we can add and IMO we should think about renaming some of the packages where the utils live but that can come separately.

adapter-cqld4/src/main/java/io/nosqlbench/engine/extensions/vectormath/CqlUtilsPluginInfo.java
jshook: We should change the packaging, post merge to match "cql utilities". The idea is to decouple across the data structure while allowing them to be complimentary.

PR #1519, opened 2023-09-07
more metric name fixes for prom expo form

### LINKED ISSUE #1517: Remove remaining invalid metric names
There are still some residual metric label constructions which contain '.' or other invalid characters. These should be removed and replaced with valid prometheus exposition format names according to that spec.

jshook:
fixes #1517


dave2wave (approved):
yes, let's underscore these!

PR #1520, opened 2023-09-07
remove result size from base op

### LINKED ISSUE #1518: Remove the base op version of result size metric
We require the result size to be present in the base op when there are specific kinds of operations which may not provide it. This is not useful and leads to API ambiguity. It should be removed.

jshook:
fixes #1518

### DIFF
--- a/adapter-mongodb/src/main/java/io/nosqlbench/adapter/mongodb/ops/MongoDirectCommandOp.java
+++ b/adapter-mongodb/src/main/java/io/nosqlbench/adapter/mongodb/ops/MongoDirectCommandOp.java
@@ -54,8 +54,4 @@ public class MongoDirectCommandOp implements CycleOp<Document> {
         return document;
     }
 
-    @Override
-    public long getResultSize() {
-        return resultSize;
-    }
 }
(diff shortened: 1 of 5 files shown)


dave2wave (approved):
LGTM

MarkWolters (commented):
I don't see where the other classes with dependencies on this are updated:
find . -name "*.java" | xargs grep getResultSize 2>/dev/null
./adapter-dynamodb/src/main/java/io/nosqlbench/adapter/dynamodb/optypes/DDBQueryOp.java:    public long getResultSize() {
./adapter-dynamodb/src/main/java/io/nosqlbench/adapter/dynamodb/optypes/DDBGetItemOp.java:    public long getResultSize() {
./adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/optypes/Cqld4FluentGraphOp.java:    public long getResultSize() {
./adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/optypes/Cqld4ScriptGraphOp.java:    public long getResultSize() {

MarkWolters (commented):
Actually I see we are not removing the OpResultSize class, just changing Op to not implement it, so I understand why the other classes aren't modified. Disregard previous comment.

MarkWolters (approved):

PR #1523, opened 2023-09-07
Nosqlbench 1522 moremath

### LINKED ISSUE #1522: support additional relevancy metrics in computation module

jshook:
fixes #1522

### DIFF
--- a/engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
+++ b/engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
@@ -189,4 +189,33 @@ public class ComputeFunctions {
         return 2.0d * ((recallAtK * precisionAtK) / (recallAtK + precisionAtK));
     }
 
+    /**
+     * Reciprocal Rank - The multiplicative inverse of the first rank which is relevant.
+     */
+    public static double RR(long[] reference, long[] sample, int limit) {
+        int firstRank = Intersections.firstMatchingIndex(reference, sample, limit);
+        if (firstRank >= 0) {
+            return 1.0d / (firstRank+1);
+        } else {
+            return 0.0;
+        }
+    }
+
+    public static double RR(long[] reference, long[] sample) {
+        return RR(reference, sample, reference.length);
+    }
+
+    public static double RR(int[] reference, int[] sample, int limit) {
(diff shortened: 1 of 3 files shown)


MarkWolters (approved):

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
```diff
@@ -146,14 +156,66 @@ public static int[] intersection(int[] reference, int[] sample, int limit) {
...
         return Intersections.count(reference, sample, limit);
     }
 
+    public static double F1(int[] reference, int[] sample) {
+        return F1(reference, sample, reference.length);
+    }
+
+    public static double F1(int[] reference, int[] sample, int limit) {
+        double recallAtK = recall(reference, sample, limit);
+        double precisionAtK = precision(reference, sample, limit);
```
dave2wave, line 178: Is there a way to reuse prior metric results from the same observation? Or does that introduce complexity that would eliminate any perceived performance gain?

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/Intersections.java
```diff
@@ -20,6 +20,32 @@
 
 public class Intersections {
 
+    /**
+     * Return a non-negative index of the first value in the sample array which is present in the reference array,
+     * OR, return a negative number. This returns array index which start at 0, not rank, which is starts at 1.
+     */
+    public static int firstMatchingIndex(long[] reference, long[] sample, int limit) {
+        Arrays.sort(reference);
```
dave2wave, line 28: If this is almost always near the beginning and our limits are generally small then a brute force check of the reference list without sorting and binarysearch may be more efficient.

dave2wave (approved):
LGTM, but the findFirst could be better without sorting and searching and just iterating - especially if the number is almost always 1.

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/Intersections.java
jshook: Good point. We do expect it to be biased towards lower indices. Interesting optimization could happen here, but these array ops are already pretty freakin fast. Let's see if it catches on fire first.

PR #1524, opened 2023-09-07
adding metadata field parsing for vector calcs

MarkWolters:


dave2wave (approved):

jshook (approved):
Later, we'll want to adjust the package names to fall within the module naming

PR #1525, opened 2023-09-08
Nosqlbench 1522 moremath

jshook:
Adding missing MAP implementation


engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
```diff
@@ -192,7 +195,7 @@ public static double F1(long[] reference, long[] sample, int limit) {
     /**
      * Reciprocal Rank - The multiplicative inverse of the first rank which is relevant.
      */
-    public static double RR(long[] reference, long[] sample, int limit) {
+    public static double reciprocal_rank(long[] reference, long[] sample, int limit) {
```
dave2wave, line 198: I like this change

dave2wave (approved):
LGTM - Go!

jshook (commented):
Some minor issues, to be addressed in the next  PR

PR #1549, opened 2023-09-15
[DRAFT] PGVector Adaptor

dave2wave:
This is a DRAFT.

So far a new `adapter-pgvector` is created by cloning `adapter-jdbc`.

Next steps involve adding the PGVector support from https://github.com/pgvector/pgvector-java#jdbc-java

1. Add the setup for the pgvector extension.
2. For inserts and selects we need to set a special PGVector object and associate with a position in a PreparedStatement. We should take an approach similar to `adapter-pinecone`, but instead write the SQL from parameters.

### DIFF
--- a/adapter-pgvector/pom.xml
+++ b/adapter-pgvector/pom.xml
@@ -1,0 +1,81 @@
+<!--
+  ~ Copyright (c) 2023 nosqlbench
+  ~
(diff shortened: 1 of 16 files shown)


jshook (approved):
LGTM as a starting point

PR #1571, opened 2023-09-28
Jshook/buildon jdk21

jshook:
This is in preparation for moving our tool chain to J21

### DIFF
--- a/hdf-loader/pom.xml
+++ b/hdf-loader/pom.xml
@@ -32,12 +32,6 @@
 
     <name>${project.artifactId}</name>
 
-    <properties>
-        <maven.compiler.source>17</maven.compiler.source>
-        <maven.compiler.target>17</maven.compiler.target>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    </properties>
-
     <dependencies>
 
         <dependency>
--- a/mvn-defaults/pom.xml
+++ b/mvn-defaults/pom.xml
@@ -26,7 +26,7 @@
 
     <properties>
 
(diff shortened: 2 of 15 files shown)


hdf-loader/pom.xml
MarkWolters, line 1: I think we should remove this module. It was a learning experience with HDF but it has nothing to do with the rest of nosqlbench. If there is interest in keeping it I can move it to another shared repo.

mvn-defaults/pom.xml
```diff
@@ -36,9 +36,8 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
         <javadoc.name>nosqlbench</javadoc.name>
-        <jacoco.version>0.8.8</jacoco.version>
```
MarkWolters, line 39: to be added back once 21 is supported?

hdf-loader/pom.xml
jshook: If the learnings are meaningful for others in the near term, let's just move the logic into some unit tests and document it as such. If you don't feel that it is worth others seeing, then we can remove it in a separate PR.

mvn-defaults/pom.xml
jshook: I moved the version down directly where it is used. There is no point in creating indirection on all things when there is only one reference.

virtdata-lib-basics/pom.xml
MarkWolters, line 1: Are these required to move to 21 or just incidental?

MarkWolters (approved):
Looks good. Looking forward to working with 21

PR #1573, opened 2023-09-28
Remove "--docker-metrics" functionality.

yabinmeng:


engine-cli/src/main/resources/docker-metrics.md
```diff
@@ -1,5 +1,11 @@
 # docker-metrics
 
+```
+NOTE: This feature has been removed and is no longer available.
```
jshook, line 4: We should change this to "... and will be replaced by a simpler mechanism soon." or something similar. Some of our biggest users rely on this and will be expecting a replacement, which we plan to provide.

jshook (changes requested):
This is a good first step. Only one requested change before merging.

engine-cli/src/main/resources/docker-metrics.md
yabinmeng: Changed

jshook (approved):
excellent, merging

PR #1574, opened 2023-09-28
remove venice adapter unless or until it is resumed, parked under a separate branch

jshook:
This module is still active on a parked branch, but it should be removed unless or until development resumes on it.


eolivelli (approved):
+1

PR #1575, opened 2023-09-29
Jshook/structure

jshook:
Tangible examples of component nesting and the first demonstrations of practical component lifecycle patterns.


nb-api/src/main/java/io/nosqlbench/adapters/api/util/TagFilter.java
```diff
@@ -107,6 +105,12 @@ public <T extends Tagged> List<T> filter(List<T> tagged) {
             .collect(Collectors.toList());
     }
 
+    public <T extends NBLabeledElement> List<T> filterLabeled(Collection<T> labeled) {
+        return labeled.stream()
+            .filter(l -> this.matches(l.getLabels().asMap()).matched())
```
MarkWolters, line 110: It seems a little confusing to have both an instance variable and a method named filter (in addition to the streaming filter method)

nb-api/src/main/java/io/nosqlbench/adapters/api/util/TagFilter.java
```diff
@@ -156,6 +160,11 @@ public TagFilter(String filterSpec) {
         }
     }
 
+//    TODO: Implement Me!
```
MarkWolters, line 163: Does this require anything more than 
this.filter.putAll(filterSpec)?

nb-api/src/main/java/io/nosqlbench/components/NBComponentSubScope.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.nosqlbench.components;
+
+public class NBComponentSubScope implements AutoCloseable {
```
MarkWolters, line 19: Not sure I understand the purpose of this class. I see it adding in AutoCloseable but why not simply have that implemented by NBBaseComponent?

MarkWolters (approved):
Looks good. Didn't leave too many comments as we already walked through this in detail for the most part. Left a couple of comments/questions if you want to respond but nothing that's a showstopper if you want to proceed to merge.

nb-api/src/main/java/io/nosqlbench/components/NBComponentSubScope.java
jshook: We did implement NBBaseComponent as such, but it seems that this causes some pain for developers with verbose warnings from static analysis anywhere an AutoCloseable is used outside of the try with resources. We'll probably move it to a similar form outside of the official AutoCloseable type.

jshook:
superseded by further work

PR #1576, opened 2023-09-29
Fix the Prometheus Push metrics reporter authentication configuration issue

yabinmeng:
For a Victoria Metrics server without authentication enabled, the PromPushReporter configuration will end up with having an invalid Rest API bearer token header as `Bearer: null`.


jshook (approved):
Nice!

PR #1602, opened 2023-10-05
Update the built-in cql-vector workload file

yabinmeng:
Also fix 2 minor issues:
1) if the cycles parameter is not explicitly specified, it should by default set to the stride number. the current code failed to do so
2) In `ActivityExecutor`, the method of `unregisterMetrics()` may be called when the method `registerMetrics()` wasn't called earlier. This will cause NPE.


jshook (approved):
Thanks!

PR #1616, opened 2023-10-10
Fleshing out pgvector adapter

MarkWolters:
- Added update op and op dispenser. 
- Changed ops to inherit from cycleOp instead of RunnableOp so we can work with results if necessary
- Changed ExecuteOp to use execute() and ExecuteQueryOp to use executeQuery()
-


jshook (approved):
lgtm

PR #1619, opened 2023-10-12
basic sanity check on tags, no newlines or spaces

jshook:
This makes yaml editing a bit safer for some cases.


MarkWolters (approved):
awesome!

Issue #1611, opened 2023-10-06
ensure that final report on metrics is shown at end of session


jshook:
This was previously carried by a custom result type known as ExecutionMetricsResult, which was a specialization of ExecutionResult.

Now that ScenarioResult the primary result type from a scenario execution, this should probably be the basis for handing results back to the session layer or similar for reporting a scenario summary (which was where the list of metrics dumped to screen previously came from.)

Ideas from Discussion:
* There is no longer one definitive stopping point for all scenarios and activities at which time a singular report should be given.
* We do not want to maintain any _ghost_ copies of components or their decorating state, like metrics. Thus, any post-results should be buffered to an appropriate channel within the normative component life-cycle hooks.
* One implementation idea would be to provide an output buffer within the component tree, (the session by default), and any component that wants to send to such a thing would find the first one at it's or a parent's level and send via that found service directly.

MarkWolters:
code complete and pull request created

PR #1670, opened 2023-11-09
Mwolters/1611

MarkWolters:
buffer metrics that go out of scope before end of test at the session level and print to screen at end of session.

See https://github.com/nosqlbench/nosqlbench/issues/1611

### DIFF
--- a/nb-api/src/main/java/io/nosqlbench/components/NBBaseComponent.java
+++ b/nb-api/src/main/java/io/nosqlbench/components/NBBaseComponent.java
@@ -130,7 +130,7 @@ public class NBBaseComponent extends NBBaseComponentMetrics implements NBCompone
         if (parent != null) {
             parent.reportExecutionMetric(m);
         } else {
-            logger.debug("buffering metric " + m.toString() + " for final report");
+            logger.warn(() -> "Component " + this.toString() + " has a non-session null parent component");
         }
     }
 
--- a/nb-api/src/main/java/io/nosqlbench/components/NBBaseComponent.java
+++ b/nb-api/src/main/java/io/nosqlbench/components/NBBaseComponent.java
@@ -111,6 +111,7 @@ public class NBBaseComponent extends NBBaseComponentMetrics implements NBCompone
             for (NBComponent child : children) {
                 child.close();
             }
+            reportMetrics();
             teardown();
         } catch (Exception e) {
             logger.error(e);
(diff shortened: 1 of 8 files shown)


nb-api/src/main/java/io/nosqlbench/components/NBBaseComponent.java
```diff
@@ -122,6 +126,14 @@ public final void close() throws RuntimeException {
         }
     }
 
+    public void reportExecutionMetric(NBMetric m) {
+        if (parent != null) {
+            parent.reportExecutionMetric(m);
+        } else {
+            logger.debug("buffering metric " + m.toString() + " for final report");
```
jshook, line 133: This log line doesn't match with any other logic I can see here.

MarkWolters: good point. Changed it to a warn level message about encountering a null parent outside of NBSession.

nb-api/src/main/java/io/nosqlbench/components/NBBaseComponent.java
```diff
@@ -122,6 +126,14 @@ public final void close() throws RuntimeException {
...
     }
 
+    public void reportExecutionMetric(NBMetric m) {
+        if (parent != null) {
+            parent.reportExecutionMetric(m);
+        } else {
+            logger.warn(() -> "Component " + this.toString() + " has a non-session null parent component");
+        }
+    }
+
```
ShaunakDas88, line 136: This is perhaps a dumb question, apologies in advance. 

Each `NBComponent` has a single parent `NBComponent`, and the root of the tree corresponds to the Session, correct? 

If so, this is baiscally just parenting up to the root/Session, and once we reach the root/Session, we just emit the warning message?

ShaunakDas88: Never mind, just realized we override this method in `NBSession`.

ShaunakDas88:
Overall makes sense. One thing I wanted to raise for discussion (and perhaps this is a bigger discussion for elsewhere, that's fine). 

Would it make sense to also store reference to appropriate root/`NBSession` for each `NBComponent`? We currently keep references to parents. I don't know how deep we expect the tree to get, but having such would allow for skipping directly up to the `NBSession`.

MarkWolters:
@jshook it was just waiting for merge, I'll go ahead and merge it now.

PR #1685, opened 2023-12-05
Support mixed-schema KNN data set formats

MarkWolters:
https://github.com/riptano/vector-perfteam/issues/84


jshook (approved):
Looks good to me. I would like for us to find a way to pre-process predicates or at least assign predicate forms to something compatible with prepared statements in the future.

PR #1703, opened 2023-12-08
Nosqlbench 1691 dnn

jshook:
Basic functions to support the euclidean DNN case.


virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/vectors/dnn/DNN_euclidean_v.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public float[] apply(long value) {
+        if (value>= N) {
+            throw new RuntimeException("You can't generate a vector for ordinal " + value + " when your population is " + this.N);
+        }
+        float[] vector = new float[D];
+        for (int idx = 0; idx < vector.length; idx++) {
+            vector[idx]= (float)((idx+value)*scale);
```
ShaunakDas88, line 50: value + idx * scale?

PR #1763, opened 2024-01-08
Jshook/at files

### LINKED ISSUE #1762: Allow command lines to reference user-provided args from within files
To allow basic workload templates which are qualified with config parameters from an external system,
users should be able to provide a file which contains command line arguments, etc.

These can be called `argfile references`, not to be confused with the built-in support for automatically reading default
args from `$NBDIR/argsfile`.

Argfile references would occur within any valid command line as `@filename`. In this form, if the named file exists, then it's args would be injected to the current command line at that position. Lists and maps should be supported as these parameters may be injects at the top level or within an activity's parameters level.

Users should be able to assert which format the parameters will be rendered in with the defaults as:
- lists as raw lists to be injects as-is, and 
- maps as raw maps to be injected as `key=value` form.

The rendered format of each argument as it will be injected should be selectable from a set of built-in formats to support global options or activity parameter formats. If no format selector is provided then the defaults above will be used.

The reference syntax will be `@` `filename` *[* `.` `extension` *]* *[* '/' `pathpart` *]* ... *|* *[* '/' *fmt* *]*

Multiple file formats can be supported eventually but the initial implementation should be limited to the basics: raw argsfile or yaml.

If a path part is specified, then the contents of the file will be traversed to that logical position, presuming tree structure, and the values to be used will be those at that position.

## examples

- `@argsfile` - presumes the standard args file format, injecting as the default raw format
- from yaml: `@argsfile.yaml:/globals/--` - asserts that there is a named section of argsfile.yaml under `globals`, and then injects the values there using the `--` formatter. This formatter requires each entry to be a global argument or an error is thrown.
- maps: `@argsfile.yaml:/activity1/=` - asserts that the top level of argsfile.yaml is a map, and inserts all map parameters as `key=value` form (using the `=` formatter).

jshook:
at-file support

close #1762


MarkWolters (approved):
LGTM!

Issue #1572, opened 2023-09-28
Configure mvn to use selected jdk21 preview features


jshook:
Some preview features of Java 21 should be enabled for regular use in the NB project. It is understood that these may change and that some refactoring may be required on subsequent versions. Nonetheless, the benefit of these changes are significant with respect to specific features and usage patterns in the NB runtime.

These should be enabled by default as a matter of project and toolchain configuration.

- [ ] 430:	String Templates (Preview)
- [ ] 443:	Unnamed Patterns and Variables (Preview)
- [ ] 446:	Scoped Values (Preview)
- [ ] 453:	Structured Concurrency (Preview)

These features are expected to provide a meaningful improvement for users and maintainers.

## 430:	String Templates (Preview)
String templating is used pervasively within NB, in many cases with different usage patterns and implementations. These should be consolidated into a centrally maintained and tested package with a durable implementation. Having JEP 430 available as a building block would greatly improve this effort.

## 443:	Unnamed Patterns and Variables (Preview)
Pattern matching without unnamed patterns is very limited. With unnamed patterns, you can actually do partial matches, which seems essential for practical use.

## 446:	Scoped Values (Preview)
Currently, we rely (too) heavily on thread locals or singletons to share scope across a thread or group of threads. This breaks encapsulation quite often, complicates the code significantly, and interferes with sound testing principles which we rely on for core systems.

## 453:	Structured Concurrency (Preview)
This is adjunct to 446 in a way, in that it will allow us to group concurrency patterns under a single and manageable view of exception handling and lifecycle management. Without this, users will continue to get spammed by 1000 of the same error when there is a syntax error in their op template, for example.

PR #1777, opened 2024-01-16
Jdk 21 preview features

workdone0:
This PR leverages Java 21 String Templates (preview feature).
Changes include:

Utilizing String Templates to replace use of string concatenation.

Updating pom.xml to enable the preview feature in adapter-cqld4 module.

References:
[Issue addressed](https://github.com/nosqlbench/nosqlbench/issues/1572)
[String Templates (Java 21 preview)](https://openjdk.java.net/jeps/430)

### DIFF
--- a/adapter-cqld4/pom.xml
+++ b/adapter-cqld4/pom.xml
@@ -160,6 +160,21 @@
                     <compilerArgs>--enable-preview</compilerArgs>
                 </configuration>
             </plugin>
+            <plugin>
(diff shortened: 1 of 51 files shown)


jshook (changes requested):
Comments on this were added to the cognitree repo: https://github.com/cognitree/nosqlbench/pull/2

Additional work is required before this change is safe and qualified for specific benefits. We should hold off until those questions are answered.

PR #1795, opened 2024-01-22
Mwolters/update compute functions

MarkWolters:


MarkWolters:
See https://docs.google.com/spreadsheets/d/1TUCdLXYqr0_1yAUYl4RBnxO3-b7kIjI3a3idzDrdZu0/edit?usp=sharing for more details

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
```diff
@@ -43,6 +45,7 @@
  * elide duplicates internally.
  */
 public class ComputeFunctions extends NBBaseComponent {
+    private final static Logger logger = LogManager.getLogger("RUNTIME");
```
jshook, line 48: This should not be "RUNTIME" as this was intended to be for component aspects specifically. (component tree changes, eventing, etc)

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
```diff
@@ -66,14 +69,16 @@ public static double recall(long[] relevant, long[] actual) {
 
     public static double recall(long[] relevant, long[] actual, int k) {
         if (actual.length < k) {
-            throw new RuntimeException("indices fewer than limit, invalid precision computation: index count=" + actual.length + ", limit=" + k);
+            logger.warn("Returned indices fewer than limit in recall calculation: index count=" + actual.length + ", limit=" + k);
```
jshook, line 72: Is this potentially an error condition which someone would want to gate on?
If so, we might want to promote it into an exception so that the error handler can handle it generically.
If not, then we might want to turn it into a metric (count the occurences).

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
```diff
@@ -85,10 +90,11 @@ public static double precision(long[] relevant, long[] actual) {
 
     public static double precision(long[] relevant, long[] actual, int k) {
         if (actual.length < k) {
-            throw new RuntimeException("indices fewer than limit, invalid precision computation: index count=" + actual.length + ", limit=" + k);
+            logger.warn("Returned indices fewer than limit in recall calculation: index count=" + actual.length + ", limit=" + k);
```
jshook, line 93: same comment as above for all these logger.warn

jshook (approved):
Mergable as is, but some quality-of-life suggestions.
Please change the logger name as indicated here or in another PR.

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
```diff
@@ -66,14 +69,16 @@ public static double recall(long[] relevant, long[] actual) {
 
     public static double recall(long[] relevant, long[] actual, int k) {
         if (actual.length < k) {
-            throw new RuntimeException("indices fewer than limit, invalid precision computation: index count=" + actual.length + ", limit=" + k);
+            logger.warn("Returned indices fewer than limit in recall calculation: index count=" + actual.length + ", limit=" + k);
         }
-        relevant = Arrays.copyOfRange(relevant,0,k);
-        actual = Arrays.copyOfRange(actual, 0, k);
+        long divisor = Math.min(relevant.length, k);
+        int arrayLength = Math.max(relevant.length, actual.length);
```
ShaunakDas88, line 75: Per our discussions on Slack, if we get a mismatch between array length of KNN and ANN, should we throw an exception?
- if `|KNN| < |ANN|`, this is indicative of either bad KNN/ground-truth data (i.e. we need to fix our test) or the DB erroneously providing too many results
- if `|ANN| < |KNN|`, this is indicative of either bad KNN/ground-truth data (i.e. we need to fix our test) or the DB erroneously providing too few results

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
```diff
@@ -85,10 +90,11 @@ public static double precision(long[] relevant, long[] actual) {
 
     public static double precision(long[] relevant, long[] actual, int k) {
         if (actual.length < k) {
-            throw new RuntimeException("indices fewer than limit, invalid precision computation: index count=" + actual.length + ", limit=" + k);
+            logger.warn("Returned indices fewer than limit in recall calculation: index count=" + actual.length + ", limit=" + k);
         }
-        relevant = Arrays.copyOfRange(relevant,0,k);
-        actual = Arrays.copyOfRange(actual, 0, k);
+        int arrayLength = Math.max(relevant.length, actual.length);
```
ShaunakDas88, line 95: Same comment as above.

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
```diff
@@ -113,14 +119,16 @@ public static double recall(int[] relevant, int[] actual) {
 
     public static double recall(int[] relevant, int[] actual, int k) {
         if (actual.length < k) {
-            throw new RuntimeException("indices fewer than limit, invalid precision computation: index count=" + actual.length + ", limit=" + k);
+            logger.warn("Returned indices fewer than limit in recall calculation: index count=" + actual.length + ", limit=" + k);
         }
-        relevant = Arrays.copyOfRange(relevant,0,k);
-        actual = Arrays.copyOfRange(actual, 0, k);
+        long divisor = Math.min(relevant.length, k);
+        int arrayLength = Math.max(relevant.length, actual.length);
```
ShaunakDas88, line 125: same comment as above.

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
```diff
@@ -132,10 +140,11 @@ public static double precision(int[] relevant, int[] actual) {
 
     public static double precision(int[] relevant, int[] actual, int k) {
         if (actual.length < k) {
-            throw new RuntimeException("indices fewer than limit, invalid precision computation: index count=" + actual.length + ", limit=" + k);
+            logger.warn("Returned indices fewer than limit in recall calculation: index count=" + actual.length + ", limit=" + k);
         }
-        relevant = Arrays.copyOfRange(relevant,0,k);
-        actual = Arrays.copyOfRange(actual, 0, k);
+        int arrayLength = Math.max(relevant.length, actual.length);
```
ShaunakDas88, line 145: same comment as above.

ShaunakDas88 (approved):
Just a comment above what we discussed on Slack, but overall, looks good.

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
MarkWolters: removed logger in latest update, WARN replaced with appropriate Exceptions

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
MarkWolters: The actual condition is comparing lengths of actual and relevant, k is simply a parameter putting a ceiling on actual. Latest update addresses this and generates Exceptions as appropriate.

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
MarkWolters: same

engine-extensions/src/main/java/io/nosqlbench/engine/extensions/computefunctions/ComputeFunctions.java
MarkWolters: Updated in latest commit to fix this

PR #1801, opened 2024-01-31
Add DNN Angular Neighbor Computation

ShaunakDas88:
We already had `DNN_angular1_v` binding which correctly creates the vectors we discussed.

The scope of this PR:
- add `DNN_angular1_neighbors` binding, for recall computation (and corresponding `DNNAngular1NeighborsTest`)
- modify assert statements in existing `DNNAngular1VTest` so that vectors with angle zero between them are precisely those in the same residue class mod the provided modulus

### DIFF
--- a/virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/vectors/dnn/DNN_angular1_neighbors.java
+++ b/virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/vectors/dnn/DNN_angular1_neighbors.java
@@ -52,6 +52,8 @@ public class DNN_angular1_neighbors implements IntFunction<int[]> {
                 )
             );
         }
+        // need to ensure each of the modulus clusters has size >= k, so that top-k nearest neighbors don't
+        // spill to another cluster with non-zero angle
         if (k * modulus > N) {
             throw new IllegalArgumentException(
                 String.format(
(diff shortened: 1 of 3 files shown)


virtdata-lib-basics/src/main/java/io/nosqlbench/virtdata/library/basics/shared/vectors/dnn/DNN_angular1_neighbors.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.function.IntFunction;
+
+/**
+ * Compute the indices of the neighbors of a given v using DNN mapping.
+ * To avoid ambiguity on equidistant neighbors, odd neighborhood sizes are preferred.
+ */
+@ThreadSafeMapper
+@Categories(Category.experimental)
+public class DNN_angular1_neighbors implements IntFunction<int[]> {
```
jshook, line 31: We may need another version of this which supports LongFunction<int[]> so that there is a native cycle-> binding type. I think it should work already as long as there is already a LongFunction -> IntFunction adapter, but since these are type (and generic type) specific, it may just be easier to add a LongFunction variant which simply calls into this one.

jshook (approved):
LGTM, only one open question, but testing will tell us if it is needed.

MarkWolters (approved):
seconding Shooky's comment about implementing LongFunction<int[]>  to avoid having to start a binding with ToInt(). But we can do that anytime, looks fine otherwise

PR #1802, opened 2024-01-31
fix for duplicate config_cyclerate gauge when striderate is used

MarkWolters:


MarkWolters:
Updated 2/1 for more elegant and streamlined fix. Only change now is to one class, metric names determined by label set in constructor.

PR #1921, opened 2024-04-09
Fix broken --list-scenarios when multiple templates are chained together

msmygit:
See #1920 for additional details.


jshook (changes requested):
I see the subtle change here. Can you add unit tests for this to stabilize it?

msmygit:
Closing this pr for now as we will tackle it separately.

PR #1923, opened 2024-04-10
Mwolters/nbio file handling

MarkWolters:
Adds resolver for NBIO cache along with command line options to configure it. This will check for the existence of a specified file in the cache and download it from the provided url if it is not already present. It supports use of sha256 checksums for file verification.


msmygit (dismissed):
LGTM 👍🏼 Will you following up with another PR to add some documentation around how to use this at https://previewdocs.nosqlbench.io?

engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLIOptions.java
```diff
@@ -206,6 +211,11 @@ public class NBCLIOptions {
     private String metricsLabelSpec = "";
     private String wantsToCatResource = "";
     private long heartbeatIntervalMs = 10000;
+    private boolean useNBIOCache = false;
+    private boolean nbioCacheForceUpdate = false;
+    private boolean nbioCacheVerify = true;
+    private String nbioCachDir;
+    private String nbioCacheMaxRetries;
```
dave2wave, line 218: Should this be a `long`?

nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ResolverForNBIOCache.java
```diff
@@ -0,0 +1,321 @@
...
+                retries++;
+            }
+        }
+        return success;
+    }
+
+    private boolean verifyChecksum(Path cachePath, URLContent checksum) {
+        try {
+            String localChecksumStr = generateSHA256Checksum(cachePath.toString());
+            Path checksumPath = Path.of(cachePath.toString().substring(0, cachePath.toString().lastIndexOf('.')) + ".sha256");
```
dave2wave, line 111: Some checksum approaches append the `.sha256` to the whole filename without stripping the outermost suffix.

nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ResolverForNBIOCache.java
```diff
@@ -0,0 +1,321 @@
...
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    private void cleanupCache(Path cachePath) {
+        if (!cachePath.toFile().delete())
+            logger.warn(() -> "Could not delete cached file " + cachePath);
+        Path checksumPath = Path.of(cachePath.toString().substring(0, cachePath.toString().lastIndexOf('.')) + ".sha256");
```
dave2wave, line 163: Creating the `checksumPath` happens more than once it should be a method.

nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ResolverForNBIOCache.java
```diff
@@ -0,0 +1,321 @@
...
+        if (!checksumPath.toFile().delete())
+            logger.warn(() -> "Could not delete cached checksum " + checksumPath);
+    }
+
+    private Path execute(NBIOResolverConditions condition, Path cachePath, URI uri) {
+        String remoteChecksumFileStr = uri.getPath().substring(0, uri.getPath().indexOf('.')) + ".sha256";
```
dave2wave, line 169: Here it is again

nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ResolverForNBIOCache.java
```diff
@@ -0,0 +1,321 @@
...
+            logger.warn(() -> "Checksum verification is disabled, returning cached file " + cachePath);
+            return execute(NBIOResolverConditions.LOCAL_NO_VERIFY, cachePath, uri);
+        } else {
+            return execute(NBIOResolverConditions.LOCAL_VERIFY, cachePath, uri);
+        }
+
+    }
+
+    private Path getOrCreateChecksum(Path cachePath) {
+        Path checksumPath = Path.of(cachePath.toString().substring(0, cachePath.toString().lastIndexOf('.')) + ".sha256");
```
dave2wave, line 243: And again.

dave2wave (changes requested):
1. There should be a single place the checksum path is derived from the path.
2. My own personal preference is to simply append `.sha256` to the path. If we do this then 1. is not as important.

ShaunakDas88 (dismissed):

nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ResolverForNBIOCache.java
dave2wave: Suppose at some point we have `textembedding.hdf5` and `textembedding.ivec` in the nbio cache? We would have a `.sha256` filename collision.

MarkWolters:
> documentation

@msmygit I can't reply in your comment but that's a good idea. Once I get this merged I'll follow up with documentation there and example yaml

nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ResolverForNBIOCache.java
MarkWolters: yeah that's a really good point. Updated code to simply append '.sha256' to complete filename / url

dave2wave (approved):
LGTM. Thanks for the change

nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ResolverForNBIOCache.java
MarkWolters: It's somewhat trivial with the other changes but I went ahead and moved it to a method, it will at least facilitate if we ever want to change strategy here.

nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ResolverForNBIOCache.java
MarkWolters: This one is different. This isn't generating a path, it's getting a String from the remote URI. But point taken, the strategy is the same and has been changed.

nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ResolverForNBIOCache.java
MarkWolters: see above comments

jshook (approved):
A good first version!

PR #1924, opened 2024-04-11
initial version of neo4j driver adapter

ShaunakDas88:
Apologies in advance for making this PR a single commit. This adds a Neo4J Java driver adapter to NoSQLBench. There are three op types added for running Cypher queries:

* [autocommit](https://neo4j.com/docs/api/java-driver/current/org.neo4j.driver/org/neo4j/driver/async/AsyncSession.html#runAsync(java.lang.String,java.util.Map,org.neo4j.driver.TransactionConfig))
* [write_transaction](https://neo4j.com/docs/api/java-driver/current/org.neo4j.driver/org/neo4j/driver/async/AsyncSession.html#executeWriteAsync(org.neo4j.driver.async.AsyncTransactionCallback))
* [read_transaction](https://neo4j.com/docs/api/java-driver/current/org.neo4j.driver/org/neo4j/driver/async/AsyncSession.html#executeReadAsync(org.neo4j.driver.async.AsyncTransactionCallback,org.neo4j.driver.TransactionConfig))

The full workload of resetting schema, creating schema, ingesting data, running ANN searches was run against a free-tier Neo4J Aura database, and all of the above appear to be working properly. One thing to keep in mind in the future is whether we are doing what is recommended for deletion of graph nodes during schema reset (a reference to support writeup is provided in the workload yaml)

In the case that we don't have the driver configured optimally, we can add in the future.


adapter-neo4j/src/main/java/io/nosqlbench/adapter/neo4j/Neo4JAdapterUtils.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class Neo4JAdapterUtils {
+
+    /**
+     * Mask the digits in the given string with '*'
+     *
+     * @param unmasked The string to mask
+     * @return The masked string
+     */
+    protected static String maskDigits(String unmasked) {
```
msmygit, line 35: #nit I think this would be a good candidate to be extracted into some common NB utility so that all drivers can utilize it.

msmygit (approved):

adapter-neo4j/src/main/java/io/nosqlbench/adapter/neo4j/Neo4JAdapterUtils.java
```diff
@@ -0,0 +1,71 @@
...
+        assert StringUtils.isNotBlank(unmasked) && StringUtils.isNotEmpty(unmasked);
+        int inputLength = unmasked.length();
+        StringBuilder masked = new StringBuilder(inputLength);
+        for (char ch : unmasked.toCharArray()) {
+            if (Character.isDigit(ch)) {
+                masked.append("*");
+            } else {
+                masked.append(ch);
+            }
```
dave2wave, line 44: Are we only masking [0-9]?

dave2wave (approved):
Interesting - I like that this is all one PR because it can be a good example of how to create a new driver.

adapter-neo4j/src/main/java/io/nosqlbench/adapter/neo4j/Neo4JAdapterUtils.java
```diff
@@ -0,0 +1,71 @@
...
+     * - https://neo4j.com/docs/api/java-driver/current/org.neo4j.driver/org/neo4j/driver/Value.html#asObject()
+     */
+    public static Object[] getFieldForAllRecords(Record[] records, String fieldName) {
+        int n = records.length;
+        Object[] values = new Object[n];
+        int idx;
+        for (int i = 0; i < n; i++) {
+            try {
+                idx = records[i].index(fieldName);
+                values[i] = records[i].get(idx).asObject();
```
MarkWolters, line 60: If any one record does not have the required field, then the call to this method fails for all and exits the loop throwing an exception. Just verifying that this is the desired behavior as opposed to eg setting values[i] for that i to null or something and continuing iteration.

ShaunakDas88: right, this was the intended behavior for this parsing of the `Record` array. 

I figured if the user is leveraging this method, they are actually expecting the field from all `Record`s that Neo4J returned to them. So if we throw an exception upon not getting the field, and this forces the user to go back and check their data and/or data model.

Should we add a `tryToGetFieldForallRecords` method here, that just populates an index in values with `null` if it's not found?

adapter-neo4j/src/main/java/io/nosqlbench/adapter/neo4j/Neo4JAdapterUtils.java
ShaunakDas88: Yes. this was shamelessly copied from the milvus driver adapter: https://github.com/nosqlbench/nosqlbench/blob/main/adapter-milvus/src/main/java/io/nosqlbench/adapter/milvus/MilvusAdapterUtils.java#L46-L64

jshook:
For the failing test, you can remove the offending maven dependency and test. This was a step towards symbolic verification of our floating point rounding error, but it will have to be added back in a more suitable way when we need it.

ShaunakDas88:
agreed, I saw Mark already removed it in his PR

adapter-neo4j/src/main/java/io/nosqlbench/adapter/neo4j/Neo4JAdapterUtils.java
msmygit: Yes, correct

jshook:
I pushed up a set of changes to reconcile with (more recent changes to) main so it should merge cleanly. As long as it goes green, we'll merge it.

jshook (approved):
A nice addition!

PR #1932, opened 2024-04-23
cqlgen: updated grammar to support UDTs and reserved keywords as identifiers

Prodyte:
* Updated grammar to allow ANTLR lexer keywords as identifiers using [ lexer modes](https://tomassetti.me/antlr-mega-tutorial/#chapter41:~:text=of%20free%20text.-,21.%20Lexical%20Modes,-We%20are%20going), except for reserved [CQL keywords](https://cassandra.apache.org/doc/4.1/cassandra/cql/appendices.html#appendix-A), which cannot be used as identifiers.

* added a new test schema file with UDTs.


jshook (changes requested):
Please do fix the createRole and alterRole queries. I don't know if it would affect users to have these broken, but where cqlgen gets used is on larger systems where these are more likely to be encountered. We should fix them before merging.

Prodyte:
> Please do fix the createRole and alterRole queries. I don't know if it would affect users to have these broken, but where cqlgen gets used is on larger systems where these are more likely to be encountered. We should fix them before merging.

I've fixed the ROLE queries. I see that `createFunction`, `createAggregate` and `alterKeyspace` were already broken.

jshook (approved):
LGTM, Thanks for this. It is a good step towards providing a better cqlgen.

PR #1942, opened 2024-05-10
Initial Qdrant driver implementation

msmygit:
Initial Qdrant driver implementation. Below core functions are implemented,

1. Delete Collection
2. Create Collection
3. Upsert Points (into a collection)
4. Count Points (from a collection)
5. Search Points (from a collection)

Along with the Qdrant, this PR also implements the below items,

1. Enhance `--logs-max` functionality to take care of cleaning up not just `.log`, but also `*.yaml` & `*.txt` files - c64fd45f6c595967a2ce86f6b46b3cf675062c55
2. Minor updates to the Milvus/Zilliz driver - 6545156ce3ac123e31b21457e6cbc0832959c1eb
3. Improve ParsedOp error message - c81968bb719dec27a14afd17dd1733609277002c
4. Trivial typo fixes - 34177a05b95f65985adc8a45818969df457685c1

### DIFF
--- a/mvn-defaults/pom.xml
+++ b/mvn-defaults/pom.xml
@@ -41,6 +41,8 @@
 
         <PROG>nb5</PROG>
         <maven.plugin.validation>VERBOSE</maven.plugin.validation>
+
+        <jacoco.version>0.8.12</jacoco.version>
     </properties>
 
     <name>${project.artifactId}</name>
@@ -549,7 +551,7 @@
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
-                <version>0.8.10</version>
+                <version>${jacoco.version}</version>
                 <executions>
                     <execution>
                         <id>prepare-agent</id>
@@ -782,7 +784,7 @@
                 <plugin>
                     <groupId>org.jacoco</groupId>
                     <artifactId>jacoco-maven-plugin</artifactId>
-                    <version>0.8.10</version>
+                    <version>${jacoco.version}</version>
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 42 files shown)


mvn-defaults/pom.xml
```diff
@@ -549,7 +551,7 @@
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
-                <version>0.8.10</version>
+                <version>${jacoco.version}</version>
```
jshook, line 554: We've moved away from indirecting these version numbers into the same place, since it makes live editing/analysis harder, and there is already a way to centralize version numbers into one place using maven dependency management.

nb-adapters/adapter-qdrant/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java-util</artifactId>
+            <!--<version>3.25.3</version>-->
+            <!-- Trying to match https://github.com/qdrant/java-client/blob/master/build.gradle#L81 -->
+            <version>3.24.0</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
```
jshook, line 50: This dependency will make it so that we can't enable this adapter by default. Guava is not well-behaved across versions where there are version conflicts in the same artifact.

jshook (approved):
Good enough to run with for experimentation. Let's do a thorough review before calling it officially done.

nb-adapters/adapter-qdrant/pom.xml
msmygit: Yes, this driver is not enabled by default. Taken care https://github.com/nosqlbench/nosqlbench/pull/1942/files#diff-1014dc69475c4c9e5f16e809bae50284fdac951a044cb289fc04bd8415d0f51eR245 and https://github.com/nosqlbench/nosqlbench/pull/1942/files#diff-c4049021bae965d533aa8d45bc1d39e332e1e07a75a866c40214f089010fc1e7R180

PR #1946, opened 2024-05-13
Add support for reading a very large payload file (as message body) directly using S4J adapter

yabinmeng:


msmygit (approved):

jshook (approved):
I would like to see this generalized a bit more in NB elsewhere, but this is fine for local use as-is.

PR #1947, opened 2024-05-13
Qdrant driver - updates

msmygit:
- Implemented full-fledged filtering logic
- Removed maven properties versioning based on PR review feedback
- Updates to create payload index operation

### DIFF
--- a/mvn-defaults/pom.xml
+++ b/mvn-defaults/pom.xml
@@ -41,8 +41,6 @@
 
         <PROG>nb5</PROG>
         <maven.plugin.validation>VERBOSE</maven.plugin.validation>
-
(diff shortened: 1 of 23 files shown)


MarkWolters (approved):
looks good. Filter logic must have been fun to write!

mvn-defaults/pom.xml
```diff
@@ -41,8 +41,6 @@
 
         <PROG>nb5</PROG>
         <maven.plugin.validation>VERBOSE</maven.plugin.validation>
-
-        <jacoco.version>0.8.12</jacoco.version>
```
msmygit, line 45: This stemmed from https://github.com/nosqlbench/nosqlbench/pull/1942#discussion_r1597161193

PR #1950, opened 2024-05-14
Additional APIs, upgraded client

msmygit:


nb-adapters/adapter-dataapi/src/main/java/io/nosqlbench/adapter/dataapi/opdispensers/DataApiDeleteOneOpDispenser.java
```diff
@@ -61,7 +61,7 @@ private DeleteOneOptions getDeleteOneOptions(ParsedOp op, long l) {
         }
         float[] vector = getVectorFromOp(op, l);
         if (vector != null) {
-            options = options.vector(vector);
```
MarkWolters, line 64: this originally was written to handle situations where vector is not null but sort is null. with the current api change that's not possible (ie there's no longer any way to call options.vector(vector) or options.sort(vector, sort) without includling a sort object) so I think the code should reflect that, e.g.
        DeleteOneOptions options = new DeleteOneOptions();
        Sort sort = getSortFromOp(op, l);
        if (sort != null) {
            float[] vector = getVectorFromOp(op, l);
            options = (vector != null) ? options.sort(vector, sort) : options.sort(sort);
        }
        return options;

msmygit: I made a slight adjustment to the above approach where in we can get a _sort_ and _vector_ (or) we get just a _vector_. Hopefully that covers all cases here.

PR #1955, opened 2024-05-30
Mwolters/sqlite

MarkWolters:
Minimal implementation of sqlite reporter


nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/engine/metrics/reporters/SqliteReporter.java
```diff
@@ -0,0 +1,183 @@
+/*
+ * Copyright (c) 2023 nosqlbench
```
msmygit, line 2:
```suggestion
 * Copyright (c) 2024 nosqlbench
```

msmygit (approved):

Issue #1944, opened 2024-05-10
Provide a better error when users select a scenario step without setting the scenario


jshook:
When the scenario step exists under at least one named scenario, and the users doesn't provide the scenario name, they should be told that they might be looking for ....

jshook:
This is complete

PR #1960, opened 2024-06-12
Better error when users select a scenario step without setting the scenario

sahankj2000:
### Description
When the scenario step exists under at least one named scenario, and the users doesn't provide the scenario name, they should be told that they might be looking for ...

Issue: [#1944](https://github.com/nosqlbench/nosqlbench/issues/1944)

```
# before
$ java -jar nb5.jar iot2.yaml main
    6035 ERROR [main] ERRORHANDLER Unable to find named scenario 'main' in workload 'iot2.yaml', but you can pick from one of: default, astra, basic_check
    6035 ERROR [main] ERRORHANDLER for the full stack trace, run with --show-stacktraces
Scenario stopped due to error. See logs for details.

# after
$ java -jar nb5.jar iot2.yaml main
    6006 ERROR [main] ERRORHANDLER Unable to find named scenario 'main' in workload 'iot2.yaml', you might be looking for one of these scenario steps: default.main, astra.main, basic_check.main
    6007 ERROR [main] ERRORHANDLER for the full stack trace, run with --show-stacktraces
Scenario stopped due to error. See logs for details.
```


jshook (approved):
LGTM

PR #1970, opened 2024-07-01
PR: QuickFix for Nested Templates + Sanitizing some block calls

MikeYaacoubStax:
1. While trying out the latest snapshot of NB5, I noticed `--list-workloads` and `--list-scenarios` did not work anymore with the following error.

> ERRORHANDLER Error from driver or included library: Error while getting workloads:Error while scanning path '/activities/astra_kv_dapi.yaml':Index 1 out of bounds for length 1

With some light debugging I figured it had to do with the nested Templates in `astra_kv_dapi.yaml`, so I attempted a fix for the nested Templates in this PR. It seemed to have fixed the problem.

There were some (unrelated) warnings for other example scenarios that seemed incomplete, and I felt I could make some very quick changes to their yamls so that at least they wouldn't WARN as we wait to complete them.

2.  A lot of older examples had an astra specific scenario which called a `schema-astra` block that simply skipped creating a keyspace (Cannot be done via CQL). However, the `schema-astra` block name seems to have been sanitized to `schema_astra` at some point, but the call to it in the scenario wasn't sanitized, resulting in an error. So I sanitized the calls in the scenarios as well.


nb-adapters/adapter-dataapi/src/main/resources/activities/find_one_and_update.yaml
```diff
@@ -0,0 +1,4 @@
+min_version: "5.21.0"
+
+description: |
+  TBD, similar to find_one_and_delete.yaml
```
MarkWolters, line 4: good catch, thanks

nb-adapters/adapter-dataapi/src/main/resources/activities/list_collection_names.yaml
```diff
@@ -6,4 +6,4 @@ blocks:
   list_collection_names:
     ops:
       op1:
-        list_collection_names:
+        list_collection_names: "TEMPLATE(collection1)"
```
MarkWolters, line 9: list_collection_names actually takes no args, so while you are correct that the yaml is broken without having *something* after the semicolon I don't think a templated var is appropriate, there's no need for the user to pass a value in. It could just be something hard coded there.

nb-adapters/adapter-dataapi/src/main/resources/activities/list_collections.yaml
```diff
@@ -6,4 +6,4 @@ blocks:
   list_collections:
     ops:
       op1:
-        list_collections:
+        list_collections: "TEMPLATE(collection1)"
```
MarkWolters, line 9: see above comment on list_collection_names

MarkWolters (changes requested):
The substantive changes look fine as far as I can tell, although Shooky is far more conversant with the logic there. I'd just request that you remove the TEMPLATE from the indicated yaml tests. Thanks Mike!

nb-adapters/adapter-dataapi/src/main/resources/activities/list_collection_names.yaml
MikeYaacoubStax: @MarkWolters sounds good! What would you suggest is a good hard coded value to add in order to guide the user to fill the rest?

MarkWolters: @MikeYaacoubStax sorry for the delay. @jshook if you have an opinion we can go with that. To me it seems like you could just put the word "database" or something along those lines. I don't see an example of another op that takes no arguments. If we want to make it stand for something I would be ok with a TEMPLATE(database_name) or similar, I just don't think it makes sense to say list collections for a collection.

jshook: For now, I would just require a value and not care what it is. There are a couple corner cases like this which we need to handle with a better yaml form, and this can be updated when we do that.

jshook (changes requested):

nb-adapters/adapter-dataapi/src/main/resources/activities/list_collection_names.yaml
MikeYaacoubStax: All good, I just put in a hardcoded TBD

nb-adapters/adapter-dataapi/src/main/resources/activities/list_collections.yaml
MikeYaacoubStax: same as above

jshook (approved):
Thanks!

jshook (approved):

MarkWolters (approved):

Issue #1973, opened 2024-07-04
Add Support for in filter for the new data-api driver adapter


sahan-ctds:
I was working with the new data-api driver adapter in NoSQLBench and found that the `in` and `nin` filters are not yet supported by this adapter. 
```
     6247 ERROR [search_three:007] BaseOpDispenser Operation in not supported
 ```
Although not currently present in the official datastax data-api docs, I found that the `in` and `nin` filters are present in the java data-api drivers (ref: [javadoc](https://datastaxdevs.github.io/astra-db-java/latest/com/datastax/astra/client/model/Filters.html#in(java.lang.String,java.lang.Object[]))).
Support for this has to be added to be able to cover more test cases while using this driver adapter.

PR #1974, opened 2024-07-05
Adding filters for data api driver adapter

sahankj2000:
Adding the missing filters as per issue: #1973


sahan-ctds:
@jshook, @MarkWolters `not` filter has not been added in yet!!!. The `not` filter takes in one other filter as an argument so feels like it should be a conjunction, but then there will be cases like `not condition1 or not condition2` where it should go under other conjunctions. Any proposals for this?
Also **IIUC** we cant have nested conditions for this adapter yet for example: `$and($or(condition1, condition2), $or(condition3, condition4))`. Do we want to plan to cover for such cases?

sahankj2000:
Hi @jshook @MarkWolters, here`not` filter has not been added in yet!!!. The `not` filter takes in one other filter as an argument so feels like it should be a conjunction, but then there will be cases like `not condition1 or not condition2` where it should go under other conjunctions. Any proposals for this?
Also **IIUC** we cant have nested conditions for this adapter yet for example: `$and($or(condition1, condition2), $or(condition3, condition4))`. Do we want to plan to cover for such cases?

MarkWolters (approved):
I think this looks good, expands the filtering type capabilities quite a bit. It would be good to see a unit test case to verify the functionality but I don't see anything that strikes me as incorrect.

jshook (approved):
LGTM

PR #1975, opened 2024-07-09
Interpolate environment in at file processing

dave2wave:
This change allows interpolation of environment variables within at file processing.


nb-engine/nb-engine-cli/src/main/java/io/nosqlbench/engine/cli/atfiles/NBAtFile.java
```diff
@@ -167,8 +168,9 @@ private static LinkedList<String> interposePath(LinkedList<String> formatted, Pa
         while (iter.hasNext()) {
             String word = iter.next();
             String modified = word.replaceAll("\\$\\{DIR}",parent.toString());
+            Optional<String> interpolatedString = NBEnvironment.INSTANCE.interpolate(modified);
```
jshook, line 171: Simply adding orThrow() here would address my earlier concerns. If there is an ENV reference which fails to lookup, we should treat it as a hard error.

jshook (changes requested):
Let's ensure that a failed env lookup throws an error.

nb-engine/nb-engine-cli/src/main/java/io/nosqlbench/engine/cli/atfiles/NBAtFile.java
dave2wave: Working on it.

MarkWolters (approved):

PR #1977, opened 2024-07-17
Cqlgen: Add conf 'disable_if_not_exists' to disable/enable 'if not exists' for DDL statements

### LINKED ISSUE #817: cqlgen should allow 'IF NOT EXISTS' for all DDL statements
For statements that are generated by the cqlgen app, it will help users avoid nuisance errors if they have 'IF NOT EXISTS' by default. This should be configurable as well, so that users can disable it when needed.

Prodyte:
Added default `IF NOT EXISTS` in these generated statements:
* create keypace
* create table
* create type

which can be disabled by setting `disable_if_not_exists: true` in cqlgen.conf.

closes #817

### DIFF
--- a/nb-adapters/adapter-cqld4/src/main/java/io/nosqlbench/cqlgen/core/CGWorkloadExporter.java
+++ b/nb-adapters/adapter-cqld4/src/main/java/io/nosqlbench/cqlgen/core/CGWorkloadExporter.java
@@ -72,7 +72,7 @@ public class CGWorkloadExporter implements BundledApp {
     private String namingTemplate;
     private double partitionMultiplier;
     private int quantizerDigits;
-    private boolean disableIfNotExists = false;
+    private boolean enableIfExists = true;
     private Map<String, List<String>> blockplan = Map.of();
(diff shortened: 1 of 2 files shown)


jshook (changes requested):
Can this be expressed in the affirmative rather than as a disable? For example, where the substitution happens, the token should be replace with "IF NOT EXISTS" IFF the boolean "enableifExists" is true, and with "" otherwise. This would more direct and simpler to follow.

Prodyte:
> Can this be expressed in the affirmative rather than as a disable? For example, where the substitution happens, the token should be replace with "IF NOT EXISTS" IFF the boolean "enableifExists" is true, and with "" otherwise. This would more direct and simpler to follow.

Done. Added `IF EXISTS` for drop statements also.

jshook (approved):
Looks better. I would align the boolean names to match the semantics of the text. On second look, "ifNotExistsEnabled" would fit the "if not exists" clause better. I'll approve this for now to unblock, but please consider this change ^

PR #1992, opened 2024-07-26
Enable java 21 preview features

### LINKED ISSUE #1572: Configure mvn to use selected jdk21 preview features

dave2wave:
closes #1572 

This adjusts the compiler args to fully enable preview features in both compiling nosqlbench and in the unit tests

### DIFF
--- a/mvn-defaults/pom.xml
+++ b/mvn-defaults/pom.xml
@@ -558,7 +558,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <!--   <argLine>-ea @{argLine}</argLine> -->
-		    <argLine>--enable-preview @{argLine}</argLine>
+                    <argLine>--enable-preview @{argLine}</argLine>
                     <systemPropertyVariables>
                         <Log4jContextSelector>
                             org.apache.logging.log4j.core.async.AsyncLoggerContextSelector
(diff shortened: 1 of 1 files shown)


jshook (approved):
LGTM

PR #1996, opened 2024-07-26
nosqlbench-1968 Add warning for --use-nbio-cache

### LINKED ISSUE #1968: Add warning for --use-nbio-cache
The --use-nbio-cache option is no longer available as caching is enabled by default. We should add a warning rather than error when the option is specified.

The intent of this is to ensure that users are not confused when they specify --use-nbio-cache. When this occurs, the user should be informed that the option has been removed and that nbio caching is enabled by default.

jshook:
closes #1968


dave2wave (approved):
LGTM

Issue #1998, opened 2024-07-26
Milvus Adapter uses a version of Log4J that has not been maintained for more than a decade


dave2wave:
I noted that `adpater-milvus/pom.xml` is using Apache Log4j 1.2.17 which is very old and not maintained. We should switch to https://reload4j.qos.ch/ which is a drop in replacement.

dave2wave:
```
<!-- https://mvnrepository.com/artifact/ch.qos.reload4j/reload4j -->
<dependency>
    <groupId>ch.qos.reload4j</groupId>
    <artifactId>reload4j</artifactId>
    <version>1.2.25</version>
</dependency>
```

Issue #191, opened 2020-08-12
Logging runtime needs updated


jshook:
Logging needs to be consolidated to a single runtime implementation.
- SLF4J needs to be removed.
- Logback needs to be removed.
- Log4J 2 needs to be used throughout.
- Programmatic constructions on Logback need to be reimplemented with Log4j
- Parent pom needs to explicitly pull in logging dependencies
- All depending modules need to remove logging configurations.
- All body arguments for calls in log4j should be converted to lambdas and tested for performance changes

dave2wave:
This appears to have been partially completed and may be to the point of mvn pom.xml file cleanup.

PR #1999, opened 2024-07-27
Maven pom file changes

dave2wave:
Small changes yet local test failures

### DIFF
--- a/nb-adapters/adapter-milvus/pom.xml
+++ b/nb-adapters/adapter-milvus/pom.xml
@@ -64,14 +64,22 @@
             <artifactId>milvus-sdk-java</artifactId>
             <version>2.4.1</version>
         </dependency>
+	<!--- CONCERNING DEPENDENCY
         <dependency>
             <groupId>log4j</groupId>
             <artifactId>log4j</artifactId>
             <version>1.2.17</version>
-        </dependency>
-        <dependency>
+        </dependency> -->
+        <!-- https://mvnrepository.com/artifact/ch.qos.reload4j/reload4j -->
+	<dependency>
+	  <groupId>ch.qos.reload4j</groupId>
+	  <artifactId>reload4j</artifactId>
+	  <version>1.2.25</version>
+	</dependency>
+	<dependency>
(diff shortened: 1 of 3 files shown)


dave2wave:
closes #1998 and #191

dave2wave:
Fixes the local test failure.

msmygit (dismissed):

nb-adapters/adapter-milvus/pom.xml
```diff
@@ -52,26 +52,21 @@
...
             <groupId>io.milvus</groupId>
             <artifactId>milvus-sdk-java</artifactId>
             <version>2.4.1</version>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/ch.qos.reload4j/reload4j replaces log4j 1.X-->
         <dependency>
-            <groupId>log4j</groupId>
-            <artifactId>log4j</artifactId>
-            <version>1.2.17</version>
+            <groupId>ch.qos.reload4j</groupId>
```
jshook, line 62: We don't want to add another logging dependency here if we can help it.
The interaction between all the modules regarding logging is complex. The strategy for managing it thus far is to 1) Standardize as much as possible on log4j2.
2) Add whatever slf4j stubs or bridging modules needed to handle per-module dependencies.

I'll look closer at this module's dependencies and understand it before making further recommendations.

dave2wave: Milvus is using the ancient unsupported Log4j 1.2.X and ch.qos.reload4j is a maintained fork of Log4J 1.2.X maintained by the original log4j author. Talk to Milvus about why they are doing this.

dave2wave: I'll provide a `mvn dependency:tree` comparison

dave2wave: It look like further study of https://www.slf4j.org/manual.html maybe required, but mostly because both cases have this:

jshook (approved):
After your verification of dependency changes, LGTM

msmygit (commented):
lgtm too!

PR #2010, opened 2024-08-01
Jshook/httpmetrics

jshook:
This attaches http metrics for status codes directly to the http adapter instead of the space, so that many spaces can co-exist without munging metric values.

It contains a few other misc improvements too.


dave2wave (approved):
LGTM - Unit tests pass. I'll make additions to use `--enable-preview` in other locations. Do we need a release note about that?

jshook:
@dave2wave I do think the error thrown by the JVM is pretty clear when that option is needed, but it wouldn't hurt to include a note about it in the release notes on the next one.

PR #2018, opened 2024-09-11
At file environment variables with default values

### LINKED ISSUE #2017: Add Default Value Support to ${env:value} At File Syntax.
**What is the ideal scenario for using this feature?**

In an atfile we want to be able to override the target, but also have the current existing value be used as a default.

**Is your feature request related to a problem? Please describe.**

Need to be more flexible with atfiles and default values.

**Would you like to help implementing or testing it?**

No.

**Additional context**

Add Unit Tests for atfile tests

dave2wave:
closes #2017


jshook (changes requested):
Adding a single character to the regex will make this syntax exactly as used in bash for similar purpose:
```
${parameter:-word}
Use Default Values. If parameter is unset or null, the expansion of word is substituted. Otherwise, the value of parameter is substituted.
```

Based on previous conversations, I think we should do that. (':-' instead of ':')

jshook (approved):
LGTM, thanks for the change!

Issue #16, opened 2020-02-26
Version number should be added to logs and possibly error messages


jshook:
When diagnosing an issue, it is sometimes difficult to verify whether an issue should have been fixed or not if the diagnostic output doesn't indicate the version. It would be much easier if all diagnostic logs had an version included, like this:

"diagnostic error line here .... (version 3.12.43)"

Steps to implement:
- [ ] Move version detection earlier in NBCLI
- [ ] Update NBCLIErrorHandler to add details when called

dave2wave:
This looks very easy. Will pass the version [here](https://github.com/nosqlbench/nosqlbench/blob/main/nb-engine/nb-engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLI.java#L137)

PR #2048, opened 2024-10-15
Store version in NBCLI class and display on some log messages

dave2wave:
Solves #16 

How to display the version was subject to interpretation. Please suggest improvements.

### DIFF
--- a/nb-engine/nb-engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLI.java
+++ b/nb-engine/nb-engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLI.java
@@ -76,8 +76,10 @@ public class NBCLI implements Function<String[], Integer>, NBLabeledElement {
     private static final int EXIT_OK = 0;
     private static final int EXIT_WARNING = 1;
     private static final int EXIT_ERROR = 2;
+    private static final String version;
 
     static {
+	version = new VersionInfo().getVersion();
         loggerConfig = new NBLoggerConfig();
         ConfigurationFactory.setConfigurationFactory(NBCLI.loggerConfig);
     }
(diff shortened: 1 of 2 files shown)


nb-engine/nb-engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLI.java
```diff
@@ -76,8 +76,10 @@ public class NBCLI implements Function<String[], Integer>, NBLabeledElement {
     private static final int EXIT_OK = 0;
     private static final int EXIT_WARNING = 1;
     private static final int EXIT_ERROR = 2;
+    private static final String version;
 
     static {
+        version = new VersionInfo().getVersion();
```
jshook, line 82: This needs to be moved below the logger config. The logger config is very sensitive to initialization order since hoisting any lo4j2 via the class loader can cause it to auto-config outside of the intended settings below.

jshook (changes requested):
Only one urgent fix. The version format that is native to the artifact coordinates works for me.

dave2wave:
> Only one urgent fix. The version format that is native to the artifact coordinates works for me.

The `getVersion()` call is identical to what is done to generate the artifact coordinates. I'm not sure what else you are asking to change?

jshook:
Let me clarify:
* The format is not that consequential. The fact that it is identical to the internal coordinate format works for me!
* There is another comment which is an urgent request for this PR, which is to fix the order of initialization in the static block.

dave2wave:
> * There is another comment which is an urgent request for this PR, which is to fix the order of initialization in the static block.

Did you look? Because that request was resolved?

jshook (approved):
LGTM

PR #2067, opened 2024-11-04
jshook/nosqlbench-2057-flattenop

### LINKED ISSUE #2057: elide RunnableOp, and flatten Op API to only CycleOp<R>
There are many cases within the NB codebase which allow the developer to distinguish between an operation which produces a result and one that doesn't. Presently these introduce many branches which are ineffectual from a performance viewpoint.

As a simplification, all occurences of Op, CycleOp, and RunnableOp can be replaced with a simple LongFunction<Result>, as encoded by the CycleOp<R> type.

In places where an operation is not expected to produce a result, the standard generic placeholder `Void` can be used. The type elision and dispatch logic should serve better than explicit branching in any case.

jshook:
closes #2057
This removes the variant types of Op and replaces them with a single function signature


dave2wave (approved):
LGTM after review and changes.

dave2wave (approved):
Still LGTM

PR #2069, opened 2024-11-05
jshook/nosqlbench-2068-opspaces

### LINKED ISSUE #2068: BUG: contextual space cache is bypassed in previous caller paths
The space cache which is used to initialize op via mappers and dispensers needs to limit access so that only the contextual version is called, which includes the cycle->index mapping logic that is op dependent.

jshook:
closes #2068

This change set fixes a hole in the space caching APIs.
Since space bindings can be set at the op-template level, any code which was bypassing this was canonically wrong. They were either 1) getting a static space before op synthesis or 2) using the space cache directly (without the name mapping layer).

The API was modified to disable bypassing the cache binding, and all clients were adjusted to use the updated path.

There are some changes in the streaming adapters which were not using the space binding rules as expected, so these were updated. These changes will need to be further reviewed with streaming devs before they use them for testing in newer NB versions.

### DIFF
--- a/nb-adapters/adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/S4JDriverAdapter.java
+++ b/nb-adapters/adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/S4JDriverAdapter.java
@@ -42,9 +42,7 @@ public class S4JDriverAdapter extends BaseDriverAdapter<S4JOp, S4JSpace> {
 
     @Override
     public OpMapper<S4JOp,S4JSpace> getOpMapper() {
-        ConcurrentSpaceCache<? extends S4JSpace> spaceCache = getSpaceCache();
-        NBConfiguration adapterConfig = getConfiguration();
-        return new S4JOpMapper(this, adapterConfig, spaceCache);
+        return new S4JOpMapper(this);
     }
 
     @Override
--- a/nb-adapters/adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/dispensers/MessageConsumerOpDispenser.java
+++ b/nb-adapters/adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/dispensers/MessageConsumerOpDispenser.java
@@ -16,6 +16,7 @@
 
 package io.nosqlbench.adapter.s4j.dispensers;
 
+import io.nosqlbench.adapter.s4j.S4JDriverAdapter;
 import io.nosqlbench.adapter.s4j.S4JSpace;
 import io.nosqlbench.adapter.s4j.ops.MessageConsumerOp;
 import io.nosqlbench.adapter.s4j.util.S4JAdapterUtil;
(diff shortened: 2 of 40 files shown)


MarkWolters (approved):
saw the changes and understood what you were doing but not *why*, until I read the PR description. Thanks for the explanation!

nb-adapters/adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/S4JDriverAdapter.java
```diff
@@ -42,9 +42,7 @@ public S4JDriverAdapter(NBComponent parentComponent, NBLabels labels) {
 
     @Override
     public OpMapper<S4JOp,S4JSpace> getOpMapper() {
-        ConcurrentSpaceCache<? extends S4JSpace> spaceCache = getSpaceCache();
-        NBConfiguration adapterConfig = getConfiguration();
```
dave2wave, line 46: This is the weird part with S4J

nb-adapters/adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/dispensers/MessageConsumerOpDispenser.java
```diff
@@ -109,28 +111,33 @@ public MessageConsumerOpDispenser(DriverAdapter adapter,
...
         HashMap<String, String> stmtLvlConsumerConfRawMap = new HashMap<>();
         for (String confKey : stmtLvlConsumerConfKeyNameList ) {
             String confVal = parsedOp.getStaticConfigOr(confKey, "");
             stmtLvlConsumerConfRawMap.put(
                 StringUtils.substringAfter(confKey, "consumer."),
                 confVal);
         }
-
-        this.combinedS4jConfigObjMap.putAll(
-            s4jSpace.getS4JClientConf().mergeExtraConsumerConfig(stmtLvlConsumerConfRawMap));
```
dave2wave, line 121: There is some weird configuration happening here because it's JMS with added Pulsar configuration

nb-adapters/adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/dispensers/MessageConsumerOpDispenser.java
```diff
@@ -109,28 +111,33 @@ public MessageConsumerOpDispenser(DriverAdapter adapter,
...
+
         S4JJMSContextWrapper s4JJMSContextWrapper = getS4jJmsContextWrapper(cycle, this.combinedS4jConfigObjMap);
         JMSContext jmsContext = s4JJMSContextWrapper.getJmsContext();
         boolean commitTransact = super.commitTransaction(txnBatchNum, jmsContext.getSessionMode(), cycle);
 
         Destination destination;
         try {
             destination = getJmsDestination(
-                s4JJMSContextWrapper, temporaryDest, destType, destNameStrFunc.apply(cycle));
+                cycle, s4JJMSContextWrapper, temporaryDest, destType, destNameStrFunc.apply(cycle));
```
dave2wave, line 140: I don't see a corresponding change to getJmsDestination

nb-adapters/adapter-s4j/src/main/java/io/nosqlbench/adapter/s4j/dispensers/MessageConsumerOpDispenser.java
```diff
@@ -139,6 +146,7 @@ public MessageConsumerOp getOp(long cycle) {
         JMSConsumer jmsConsumer;
         try {
             jmsConsumer = getJmsConsumer(
+                cycle,
```
dave2wave, line 149: Is there a change to getJmsConsumer

dave2wave (approved):
I think that this can be applied and then the preview build should be tested with s4j, pulsar, and kafka adapters a;pong with the Streaming team.

PR #2070, opened 2024-11-05
Add NBAdvisor Framework

### LINKED ISSUE #12: Implement NBAdvisor layer for all detectable issues which may cause unexpected behavior.
Assertive mode is a feedback setting which will tell novice users when something doesn't look right for a user-interactive scenario.

Examples:

There are zero cycles in an activity
There are zero activites in a scenario
a parameter isn't being used
In some cases, especially with automated testing calling into eb, it is valid for these conditions to occur. For those cases, it would be nice to be able to say assertive=false to delegate trust to the caller for configuration sanity.

Originally filed as https://github.com/engineblock/engineblock/issues/184

The steps to implement this are:
- [ ] Add enum for the different validation levels.
- [ ] Document validation level semantics.
- [ ] Enable validation for unused parameters
- [ ] Enable validation for client resource overloads.

After reviewing other issues related to this, it seems that there are some distinct level of feedback that are appropriate:
1. ignore - This means provide no user feedback and allow all settings to be applied as given if possible.
2. advise - Inform the user of cases which appear to be problematic or which could cause unexepected behavior.
3. dryrun - Exercise the advisory logic, but use dryrun semantics where all advise is accumulated into a single output that the user can cross-check things with.
4. enforce - Throw an error in any case which advise would normally alert the user.

Essentially in every case where advise would cause user feedback, enforce should throw an error. In both cases, they same advisory text should be given about the potential bad effects. And if possible, example syntax for an alternate way of addressing the issue should be provided. Ideally, such example syntax is usable copypasta for the current scenario.

For implementing this, a new type called NBAdvisor should be created which enforces a few contractual requirement for developers. Specifically developers should be required to construct valid advisor objects which:
1. Knows its usage context and how to log or reference docs using some distinct and stable name.
2. Has a uniform output which can be marshalled into some advisor specific view, such as with a dry-run test output.
3. Provides details which clearly explain why the advise is being given.
4. If possible provide an example of a copypasta which can be used to fix the user's current configuration.

dave2wave:
Closes #12


nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityimpl/uniform/BaseDriverAdapter.java
```diff
@@ -180,8 +180,10 @@ public NBConfigModel getConfigModel() {
             .add(Param.optional(List.of("workload", "yaml"), String.class, "location of workload yaml file"))
             .add(Param.optional("driver", String.class))
             .add(Param.defaultTo("dryrun", "none").setRegex("(op|jsonnet|emit|none)"))
+            .add(Param.defaultTo("advisor", "none").setRegex("(validate|enforce|none)"))
```
jshook, line 183: There are other commands besides start and run which may need advisor. We might want to hoist this parameter up to NBCLI level.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/errors/ProcessingEarlyExit.java
```diff
@@ -0,0 +1,40 @@
...
+    public ProcessingEarlyExit(String message, int exitCode) {
+        this.message = message;
+        this.exitCode = exitCode;
+    }
+
+    public String toString() {
+        final String helpmsg = """
+                'ARG'.
+                """
+            .replaceAll("ARG", this.message);
```
jshook, line 33: Consider using STR templates for this kind of message.

nb-engine/nb-engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLI.java
```diff
@@ -136,12 +136,12 @@ public Integer apply(final String[] args) {
                     break;
                 }
 
-            final String error = NBCLIErrorHandler.handle(e, showStackTraces, NBCLI.version);
+            final int result = NBCLIErrorHandler.handle(e, showStackTraces, NBCLI.version);
             // Commented for now, as the above handler should do everything needed.
-            if (null != error) System.err.println("Scenario stopped due to error. See logs for details.");
```
jshook, line 141: It seems like telling the user there was an error is important. Why remove this detail?

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/process/NBCLIErrorHandler.java
```diff
@@ -43,7 +44,7 @@ public class NBCLIErrorHandler {
 
     private final static Logger logger = LogManager.getLogger("ERRORHANDLER");
 
-    public static String handle(Throwable t, boolean wantsStackTraces, String version) {
+    public static int handle(Throwable t, boolean wantsStackTraces, String version) {
```
jshook, line 47: I'm not sure what the benefit of this change is except it makes this class more "posix"

jshook (dismissed):
Some minor comments/questions, but no blockers.

nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityimpl/uniform/BaseDriverAdapter.java
dave2wave: The NBCLI level differs. We should discuss.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/errors/ProcessingEarlyExit.java
dave2wave: I should redo this as I was copying a pattern that no longer makes sense.

nb-engine/nb-engine-cli/src/main/java/io/nosqlbench/engine/cli/NBCLI.java
dave2wave: I can adjust this.

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/core/lifecycle/process/NBCLIErrorHandler.java
dave2wave: It's due to the new early exit handling which is either an error or not.

nb-apis/nb-api/pom.xml
```diff
@@ -161,6 +161,12 @@
             <artifactId>sqlite-jdbc</artifactId>
             <version>3.46.0.0</version>
         </dependency>
+      <dependency>
+        <groupId>io.qdrant</groupId>
+        <artifactId>client</artifactId>
+        <version>1.9.1</version>
+        <scope>test</scope>
+      </dependency>
```
jshook, line 169: This does not belong in an API module. Any vendor or driver-specific dependencies need to be in their respective modules.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/advisor/NBAdvisorResults.java
```diff
@@ -0,0 +1,76 @@
...
+        List<NBAdvisorPoint.Result<?>> results = getAdvisorResults();
+        Iterator<NBAdvisorPoint.Result<?>> iterator = results.iterator();
+        int count = 0;
+        boolean terminate = false;
+        Level level = Level.INFO;
+        while (iterator.hasNext()) {
+            NBAdvisorPoint.Result<?> result = iterator.next();
+            level = result.isError() ? result.conditionLevel() : level.INFO;
+            switch (NBAdvisorLevel.get()) {
+            case NBAdvisorLevel.none:
```
jshook, line 49: Formatting is still off on these. I can do a reformat commit if you want.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/advisor/NBAdvisorResults.java
```diff
@@ -0,0 +1,76 @@
...
+            case NBAdvisorLevel.validate:
```
jshook, line 50: Is it intended that the "none" case falls through to the "validate" case? Without a break after none, this will happen.

nb-apis/nb-api/pom.xml
dave2wave: This came from one of your changes

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/advisor/NBAdvisorResults.java
```diff
@@ -0,0 +1,76 @@
...
+                if ( level == Level.ERROR ) {
+                    System.out.println(result.rendered());
```
jshook, line 52: Output should go through the system logger as a standard mechanism. (Console is wired for this)

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/advisor/NBAdvisorResults.java
dave2wave: Please

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/advisor/NBAdvisorResults.java
```diff
@@ -0,0 +1,76 @@
...
+                }
+                break;
+            }
+        }
+        if ( terminate ) {
+            String message = String.format("Advisor found %d actionable %s.",
+                    count,
+                    (count < 2 ? "error" : "errors"));
+            System.out.println(message);
+            throw new ProcessingEarlyExit(message, 2);
```
jshook, line 71: I would like for the name of ProcessingEarlyExit to be more descriptive, as in whether it is expected or not. I know it means "not expected", but it would be nice for it to be more clear as to what/why it is. Maybe something like "AdvisorErrorCondition" or similar.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/advisor/NBAdvisorResults.java
dave2wave: yes and yes. For now I'm not displaying every test, but only the "ERRORS".

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/advisor/NBAdvisorResults.java
dave2wave: I'll switch to the Console Logger Monday, or feel free to do so.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/advisor/NBAdvisorResults.java
dave2wave: Sure. A name change is fine.

nb-apis/nb-api/src/test/java/io/nosqlbench/nb/api/components/core/NBComponentServicesTest.java
```diff
@@ -65,6 +66,7 @@ public void testComponentServices() {
             System.out.println("metric: " + m.toString());
         });
 
+        //throw new RuntimeException("implement advisor tests here");
```
jshook, line 69: remove this?

jshook (changes requested):
Only a couple key changes needed, but several suggestions to consider too

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/advisor/NBAdvisorResults.java
dave2wave: done

nb-apis/nb-api/src/test/java/io/nosqlbench/nb/api/components/core/NBComponentServicesTest.java
dave2wave: done

jshook (approved):
LGTM, happy to see this in the code!

Issue #807, opened 2022-11-30
Having an empty scenario causes an error


msmygit:
**Describe the bug**
When we have an empty scenario block similar to this in the workload yaml file. Even if we call another scenario block (e.g. `block1` in this case) `...tags=block:block1...`:
```
scenarios:
  default:
  block1:
     ...
```
which gets an error like as below:
```
1589 ERROR [main] ERRORHANDLER cause (see stack trace for details):javax.script.ScriptException: io.nosqlbench.api.errors.BasicError: Unable to use a null value for scenario named default in yaml.
java.lang.RuntimeException: javax.script.ScriptException: io.nosqlbench.api.errors.BasicError: Unable to use a null value for scenario named default in yaml.
```
Reproduced this with `4.17.29`.
**To Reproduce**
See above

**What was Expected**
It should not refer to the unspecified empty scenario block (`default` in this case) and called the `block1`.

**Additional context**
OS: macOS 12.6.1 Monteray
environment: n/a
version info (`./nb --version` or `java -jar nb.jar --version`): `4.17.29`

**Screenshots, if applicable**
n/a

PR #2078, opened 2024-11-13
Allow Empty Scenarios

dave2wave:
closes: #807


msmygit (approved):

jshook (approved):
There was a user-impacting reason for the null scenario check to be added before, but I think it was lost. I'm ok with reverting it for now. We'll adjust if we see the original issue again.

Issue #442, opened 2022-02-24
Remove ActivityType interface e:2d


jshook:
Once all drivers are converted to the DriverAdapter API, the ActivityType API needs to be removed as a separate type.

The goal of this change is to reduce ambiguity in the developer-facing APIs. If the ActivityType interface is no longer implementable, the developers will understand that the only valid extension point for this type of feature is the DriverAdapter.

dave2wave:
AFAICT - This ActivityType class code is still present.

dave2wave:
The ActivityType is almost unwound.

dave2wave:
To remove `ActivityType` the use of `StandardAdapterType` must be removed from `ActivityTypeLoader` and `ActivityLoader`.

dave2wave:
Current use is around Alias resolultion - `ActivityLoader`
```
public class ActivityLoader {
    private static final Logger logger = LogManager.getLogger("ACTIVITIES");
    private final Map<String, Activity> activityMap = new ConcurrentHashMap<>();

    public ActivityLoader() {
    }

    public synchronized Activity loadActivity(ActivityDef activityDef, final NBComponent parent) {
        activityDef= activityDef.deprecate("yaml","workload").deprecate("type","driver");
        final Activity activity = new StandardActivityType<>(activityDef, parent).getAssembledActivity(activityDef, this.activityMap, parent);
        this.activityMap.put(activity.getAlias(),activity);
        ActivityLoader.logger.debug("Resolved activity for alias '{}'", activityDef.getAlias());
        return activity;
    }

    public void purgeActivity(final String activityAlias) {
        activityMap.remove(activityAlias);
    }
}
```

dave2wave:
But the `ActivityLoader` is core within `ContainerActivitiesController`

```
    private ActivityRuntimeInfo doStartActivity(ActivityDef activityDef) {
        if (!this.activityInfoMap.containsKey(activityDef.getAlias())) {
            Activity activity = this.activityLoader.loadActivity(activityDef, this);
            activity.initActivity();
            ActivityExecutor executor = new ActivityExecutor(activity);
            Future<ExecutionResult> startedActivity = executorService.submit(executor);
            ActivityRuntimeInfo activityRuntimeInfo = new ActivityRuntimeInfo(activity, startedActivity, executor);
            activityRuntimeInfo.getActivityExecutor().awaitMotorsRunningOrTerminalState();
[...]

PR #2079, opened 2024-11-14
Cleanup for Deprecated ActivityType

dave2wave:
most solves: #442

### DIFF
--- a/nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityapi/core/LegacyActivityType.java
+++ b/nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityapi/core/LegacyActivityType.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023 nosqlbench
+ * Copyright (c) nosqlbench
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


dave2wave:
`ActivityType` is renamed as `LegacyActivityType` and is only used by `StandardActivityType`.

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityapi/core/LegacyActivityType.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023 nosqlbench
+ * Copyright (c) nosqlbench
```
msmygit, line 2: let's leave this copyright year intact

msmygit (commented):
LGTM. I'd leave this one for @jshook to review as it touches the very core.

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityapi/core/LegacyActivityType.java
dave2wave: We've been removing. It is actually irrelevant. You can't reset the year. The year is the original year of the overall product. Copyright protections are time limited and don't reset.

jshook (commented):
I'll have a way to completely collapse this down to eliminate the extraneous types. I can tackle it next week. Given that, I'm not sure that the doc changes or naming changes are significant.

dave2wave:
I will leave this open and you can either work from this, or follow your path to fix #442

jshook (approved):
As a disambiguation, I'm ok with this

Issue #2080, opened 2024-11-14
NBCLI lists out workloads and scenarios. Add usage information.


dave2wave:
**What is the ideal scenario for using this feature?**
When listing out components in the system with variations of `--list-*` provide a more complete example of the command and parameters needed.

**Is your feature request related to a problem? Please describe.**
It's an improvement to documentation discovery

**Would you like to help implementing or testing it?**
Yes.

**Additional context**

PR #2081, opened 2024-11-14
Make listings similar to command syntax

dave2wave:
closes: #2080 

The listings for `--list-workloads` and `--list-scenarios` were subtly different from current cli syntax which made users think that the examples were missing.


dave2wave:
This needs a little more work. Leaving as a draft.

PR #2082, opened 2024-11-15
Update and align dependencies

dave2wave:
When analyzing `mvn dependency:list | sort | uniq` I noticed some version conflicts with dependencies. This PR updates several.

### DIFF
--- a/mvn-defaults/pom.xml
+++ b/mvn-defaults/pom.xml
@@ -357,16 +357,6 @@
                 <artifactId>jackson-jaxrs-json-provider</artifactId>
                 <version>2.18.1</version>
             </dependency>
-            <dependency>
-                <groupId>com.sun.xml.bind</groupId>
-                <artifactId>jaxb-core</artifactId>
-                <version>4.0.5</version>
-            </dependency>
-            <dependency>
-                <groupId>com.sun.xml.bind</groupId>
-                <artifactId>jaxb-impl</artifactId>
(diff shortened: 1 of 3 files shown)


mvn-defaults/pom.xml
```diff
@@ -355,17 +355,17 @@
...
                 <groupId>com.fasterxml.jackson.jaxrs</groupId>
                 <artifactId>jackson-jaxrs-json-provider</artifactId>
-                <version>2.9.8</version>
+                <version>2.18.1</version>
             </dependency>
             <dependency>
                 <groupId>com.sun.xml.bind</groupId>
                 <artifactId>jaxb-core</artifactId>
-                <version>2.3.0.1</version>
+                <version>4.0.5</version>
```
jshook, line 363: This seems like a big change. Sometimes these changes are not runtime compatible. Can we verify this with downstream docs somewhere, or ensure it with some additional verification.

jshook: Or even better, remove the dependency chain if it is vestigial.

jshook (changes requested):
One item for research

mvn-defaults/pom.xml
dave2wave: Removed and the build functioned.

dave2wave:
> One item for research

Research completed

jshook (approved):
Awesome change!

Issue #1515, opened 2023-09-06
Make template variables specifiable in a way which requires them to be set directly


jshook:

jshook:
Using these TEMPLATE forms in layers with other forms is not supported.

dave2wave:
In StrInterpolator:
```
        @Override
        public String lookup(String key) {
            String value = null;

            String[] parts = key.split("[:,]", 2);
            if (parts.length == 2) {
                key = parts[0];
                value = parts[1];
                // HERE
                if (!extractedDefaults.containsKey(key)) {
                    extractedDefaults.put(key,value);
                }
            }

            for (Map<String, ?> map : maps) {
                Object val = map.get(key);
                if (val != null) {
                    value = val.toString();
                    break;
                }
            }
            value = (value==null? extractedDefaults.get(key) : value);

            value = (value != null) ? value : warnPrefix + ":" + key;

//            if (accesses.containsKey(key) && !accesses.get(key).equals(value)) {                                                                   
//                throw new OpConfigError("A templated variable '" + key + "' was found with multiple default values: '" + accesses.get(key) + ", and " + value +". This is not allowed." +                                                                                                              
//                    " Template variables must resolve to a single value.");                                                                        
//            }                                                                                                                                      

            accesses.put(key,value);
            logger.trace("Template parameter '" + key + "' applied as '" + value + "'");
            // TODO summarize these to how many times                                                                                                
            return value;

        }
```

dave2wave:
substitutionType cases:
1. "" or "-": current op - use `key`'s value if present otherwise the `value`.
2. "=" : use `key`'s value - if missing set the `key`'s value to `value`.
3. "?" : use `key`'s value if present otherwise the this is an error exit condition.
4. "+": use `value` if the `key` is present otherwise ""

PR #2092, opened 2024-11-20
Commented TEMPLATEs in yamls are ignored

### LINKED ISSUE #1927: Handle commented-out template variables work as expected.
In any case, nosqlbench should understand if a template variable occurs within a commented line of YAML, so that it may provide appropriate feedback for corner cases.

This should use the validation settings from #12 

In the relaxed mode, the presence of unused (commented-out) template variables should not matter, but if there is a discrepancy, the user should get a warning.

In the assertive mode, if a template variable is commented out in every position, yet it is still provided, then this should be considered an error, and the user informed of the details. Template variables which are used in at least one position (not commented out) should not be considered an error.

dave2wave:
closes #1927

### DIFF
--- a/nb-apis/adapters-api/src/test/java/io/nosqlbench/adapters/api/templating/StrInterpolatorTest.java
+++ b/nb-apis/adapters-api/src/test/java/io/nosqlbench/adapters/api/templating/StrInterpolatorTest.java
@@ -86,6 +86,13 @@ public class StrInterpolatorTest {
     }
 
     @Test
+    public void shouldMatchWithComments() {
+        StrInterpolator interp = new StrInterpolator(abcd);
+        String a = interp.apply("TEMPLATE(start,START)\n# TEMPLATE(blahblah,blah)\nTEMPLATE(keydist,Uniform(0,1000000000)->int);");
+        assertThat(a).isEqualTo("START\n# TEMPLATE(blahblah,blah)\nUniform(0,1000000000)->int;");
+    }
+
+    @Test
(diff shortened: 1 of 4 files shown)


dave2wave:
Additonally NBAdvisor does not show before and after TEMPLATE expansion and advisorLevel = enforce will stop if the old <<key:value>>  format is used.

dave2wave:
I plan to tackle #1515 next. For that I will be changing the implementation of `apply` and `matchTemplates`.  I'm going to make this a Draft.

nb-apis/adapters-api/src/test/java/io/nosqlbench/adapters/api/templating/StrInterpolatorTest.java
```diff
@@ -85,6 +85,13 @@ public void shouldMatchNestedParens() {
         assertThat(a).isEqualTo("Uniform(0,1000000000)->int;");
     }
 
+    @Test
+    public void shouldMatchWithComments() {
+        StrInterpolator interp = new StrInterpolator(abcd);
+        String a = interp.apply("TEMPLATE(start,START)\n# TEMPLATE(blahblah,blah)\nTEMPLATE(keydist,Uniform(0,1000000000)->int);");
```
jshook, line 91: Just a note for later, you can use the triple quote format to make multi-line strings easier to see

jshook (dismissed):
looks good

jshook (approved):
This looks like a very good simplification. Maybe as a follow-on we can see how far we are from removing the upstream dependency from before.

PR #2094, opened 2024-11-20
many-sessions improvements, type fixes, more javadocs

jshook:
Key changes here:
- Improved the adapter API and updated call sites
- Added an example driver adapter
- Squelched the session leak indicator by default in CQL to support many-session testing
- Included process-level open file descriptor count in metrics

Reviewers may benefit from selecting specific commits, since the refactoring noise is isolated to a couple, and the others are self-evident


dave2wave (approved):
There are a lot of changes which fit a few patterns in the improvements.

Really like the idea of example adapter.

LGTM

PR #2096, opened 2024-11-21
Add bash style variable substitution types

### LINKED ISSUE #1515: Make template variables specifiable in a way which requires them to be set directly

dave2wave:
closes #1515

### DIFF
--- a/nb-apis/adapters-api/src/test/java/io/nosqlbench/adapters/api/templating/StrInterpolatorTest.java
+++ b/nb-apis/adapters-api/src/test/java/io/nosqlbench/adapters/api/templating/StrInterpolatorTest.java
@@ -16,6 +16,8 @@
 
 package io.nosqlbench.adapters.api.templating;
 
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
 import org.junit.jupiter.api.Test;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


nb-apis/adapters-api/src/test/java/io/nosqlbench/adapters/api/templating/StrInterpolatorTest.java
```diff
@@ -115,6 +117,30 @@ public void shouldMatchWithComments() {
...
+    public void sustitutionTests() {
+        StrInterpolator interp = new StrInterpolator(abcd);
+        String a = interp.apply("-${setkey:=setme}-${setkey}-");
+        assertThat(a).isEqualTo("-setme-setme-");
+        String b = interp.apply("-${setkey2:-setme}-${setkey2:-setyou}-");
+        assertThat(b).isEqualTo("-setme-setyou-");
+        String c = interp.apply("-${setkey:=setme}-${setkey3:+setme}-${setkey:+setyou}-");
+        assertThat(c).isEqualTo("-setme-setme-setyou-");
+    }
+
```
jshook, line 130: Can we add a check for the nested case of at least one level?

dave2wave: Sure we can add any test you want.

nb-apis/adapters-api/src/test/java/io/nosqlbench/adapters/api/templating/StrInterpolatorTest.java
jshook, line 1: If we want to make this a standard, as in verified and shown in literate testing docs, then we can add it to the bank of tests under UniformWorkloadSpecificationTest, which basically means adding a doc/example triplet under nb-apis/adapters-api/src/main/resources/workload_definition. (The conventions and format are documented there)

dave2wave: OK, new issue.

Issue #375, opened 2021-11-04
Allow NBIO paths to have qualifier prefixes


jshook:
**What is the ideal scenario for using this feature?**
A user specifies a local file, classpath resource, or a URL, and they want to be able to specific which of these locations to limit the search in, to avoid ambiguity about the source of the content.
The formats suggested are:
- `file:...` should limit the search to the default filesystem
- `cp:...` or `classpath:...` should search the classpath
- `local:...` should search both classpath and the filesystem
- `all:...` should provide the default behavior for cases where an explicitly selected behavior is needed for scripting integration and so on.
- URLs are already handled with protocol-specific handlers, and do not need separate qualifiers, as these usage scenarios are mutually exclusive. (The `all:...` scenario encapsulates this)

Further, a format of `cp,file:...` may be used to specific an alternate order of precedence.

**Is your feature request related to a problem? Please describe.**
When users have the same-named file in multiple locations, the order of precedence is not obvious.

**Would you like to help implementing or testing it?**

**Additional context**

dave2wave:
Why reinvent uri protocols? I asked ChatGPT about a http+file: and here is the reply:

The URI (Uniform Resource Identifier) protocol field does not have a standard syntax for specifying an "either/or" approach such as `http+file:`. The URI scheme is designed to indicate a single protocol or access method to be used. Here are a few points about the structure and use of URIs:

1. **Standard URI Schemes**: Each URI scheme (like `http`, `https`, `ftp`, `file`, etc.) specifies a single protocol. For example, `http://example.com` specifies the HTTP protocol, and `file:///path/to/file` specifies the file protocol.

2. **Combining Protocols**: There isn't a standard way to combine multiple protocols in a single URI scheme. The idea of `http+file:` is not recognized by any standard and would not be understood by web browsers or other URI processors.

3. **Custom Schemes**: While creating custom URI schemes (like `myapp:`) is possible, they require custom handlers to interpret them. They are not part of any official standard and would need to be explicitly supported by the application using them.

4. **Alternative Approaches**:
    - **Conditional Logic in Applications**: Instead of combining protocols in the URI, handle the protocol decision within the application logic. For instance, your application could try accessing a resource via `http` first and fall back to `file` if the HTTP request fails.
    - **Use of Proxy**: In some cases, a proxy or intermediary service could be used to determine the appropriate protocol to use based on availability or other criteria.

[...]

PR #2098, opened 2024-11-22
Add Protocol Resolver Chain Selection to NBIO

dave2wave:
closes: #375

### DIFF
--- a/nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ParseProtocol.java
+++ b/nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ParseProtocol.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) nosqlbench
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


dave2wave:
With this PR it is now possible to run both of these command lines:
1. Classpath `java --enable-preview -jar nb5.jar run workload==cp:qdrant_vectors_live.yaml`.
2. Filesystem `java --enable-preview -jar nb5.jar run workload==file:qdrant_vectors_live.yaml`.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ParseProtocol.java
```diff
@@ -0,0 +1,56 @@
...
+        if (filepath.startsWith("http")) {
+            path = filepath;
+            protocols.add("all");
+        } else {
+            String[] parts = filepath.split(":");
+
+            if (parts.length < 2) {
+                path = filepath;
+                protocols.add("all");
+            } else {
```
jshook, line 42: In some cases, I think there may be a ':' used as a suffix or qualifier, as in atfiles. There was also a previous convention used which would allow users to specify which version of yaml structure to load with `:v1` and `:v2` suffixes, but this _should_ be long gone.

I would suggest limiting the split to two terms (split method supports this), and ensuring that the first word is one from a specific set of accepted protocols, preferrably from an enum.

jshook (changes requested):
Please make the protocol parser more selective.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/nbio/ParseProtocol.java
dave2wave: I'll make this a more linear scan ...

dave2wave: Changed to enums

dave2wave:
> Please make the protocol parser more selective.

Done. It only follows it's narrow match list and if anything fails then it treat the `filename` as if it were plain.

jshook (approved):
Looks good!

PR #2101, opened 2024-11-25
StandardActivityType absorbs legacy dispenser logic

### LINKED ISSUE #442: Remove ActivityType interface e:2d

### LINKED ISSUE #443: Initiative: Uniform Driver Capabilities
[UPDATE] Moving the sub-items from this directly to the parent milestone for better visibility.

One of the key promises of NoSQLBench is that of re-usable concepts and workflows. This means that time invested by users to learn more advanced usage patterns will be rewarded with an understanding of how to apply that knowledge elsewhere. The driver interfaces and capabilities are a key focal point for this. In other words, when the drivers support a uniform set of integrations and behaviors, the toolkit becomes simpler and more powerful at the same time.

## Challenge

Previously, driver developers had to effectively copy and emulate the behavior which was established by convention, using code which was not managed by the core runtime. As well, the concepts and tooling needed to streamline the process of converting templates to operations was lacking in terms of clarity and supportive APIs.

## Remedy

1. [x] #430
3. [x] #432
4. [x] #433
8. [x] #437
10. [x] #439
11. [x] #440
12. [x] #441
13. [x] #442
14. [x] #421

dave2wave:
Closes #442
Closes #443


jshook (approved):
Woot!

Issue #1808, opened 2024-02-06
Add strict mode for well-formed names in NB identifiers


jshook:
At some point in the near future, names which can cause issue with downstream metrics systems should be
avoided completely. This means that any name which includes dashes, or otherwise does not conform to a valid open metrics label name or value should cause an error.

As the first step, places where users can provide these labels should be enabled with advisory warnings indicating that the format will be deprecated, suggesting an alternate example form.

This includes these places at least:

- [x] * workload parameters - `yaml`
- [x] * template variables - `StrInterpolator`
- [x] * label names and values
- [x] * tag names - `yaml`
- [ ] * op field properties which may be included in metrics or logging - `yaml`
- [x] * activity names, container names, container variable names - `yaml`

dave2wave:
There are quite a few locations mentioned. It would help to point to code where some of these checks could be made.

An approach to this would be to create a validator class that would be a static singleton like NBEnvironment. Then each type of name could be tested in the appropriate piece of code.

dave2wave:
@jshook does this issue depend on the completion of #12 ?

jshook:
Issue #12 is part of this, however it is mostly for the cohesion in docs. In other words, we want the behavior of this validator logic to be very well defined for users. However, we should not relax this because it is a strong requirement for sane downstream metrics. So in terms of what is validated at what level, it should at least be documented as a validation that occurs even for relaxed mode.

dave2wave:
@sahankj2000 identified the class that ought to be modified in their closed PR - `StrInterpolator`:

I think that we can consider code like this:

```java
import org.apache.commons.text.StringSubstitutor;
import org.apache.commons.text.lookup.StringLookup;

public class StringModifierWithSubstitutor {
    public static void main(String[] args) {
        String original = "example-string-to-modify";

        // Create a custom lookup to replace '-' with '_'
        StringLookup lookup = key -> key.replace("-", "_");

        // Set up the StringSubstitutor with the custom lookup
        StringSubstitutor substitutor = new StringSubstitutor(lookup);
        
        // Perform the substitution
        String modified = substitutor.replace(original);

        // Check if modification happened
        if (!original.equals(modified)) {
            System.out.println("Modification occurred. Modified string: " + modified);
        } else {
            System.out.println("No modification was needed.");
        }
    }
}
```

We will need to have this happen according to the advisory levels.

dave2wave:
`yaml` work starts from within `OpsLoader.java`
```java
[...]

Issue #1609, opened 2023-10-06
make activityDef and activityDef.getParams consistent across all standard activity params


jshook:
example: setting "cycles" on the backing store does not properly update the spec models at the activityDef

workdone0:
@jshook Can you tell me how to reproduce this issue?

jshook:
This is related to the way we track activity parameters and expose them for changes dynamically while an activity runs.
In nb 5.21+, this is handled with component tree propogation of change event types ideally. Further, each of these parameters which is part of some parsed state or stateful view needs to use the same reference to the active value, no matter whether it is a primitive or another type. 

Once we have made this work well for cycles (assuming the issue above is still present) it should be documented as a pattern in the dev guidef or dynamic activity parameters.

@workdone0 I think this particular issue may have already been solved. We need to try to reproduce it by setting the parameter on the ActivityDef while an activity is running, and then asserting that the cycles model on that type synchronously updated and consistent. The two places which could get out of sync would be the string version of the cycles spec on the ActivityDef params map, and the Cycles spec model which is also on that type.

dave2wave:
Include in the parsing of ActivityDef to Params conversion to proper lowercase with _ format.

dave2wave:
In `nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/config/params/ParamsParser.java` the current rule for a valid name is as follows:

```
 * <h3>Names</h3>                                                                                                                                                                      
 * The name of a parameter must appear in raw literal form, with no escaping. A name may not                                                                                           
 * contain spaces, semicolons or equals signs, and may not start with quotes of any kind.                                                                                              
 * Otherwise, there are no restrictions placed on the characters that may appear in                                                                                                    
 * a name. Even escaped characters are allowed.                                                                                                                                        
```

This is the code that parses ActivityDefs. It looks like one spot to enforce more limited character set rules. There are two approaches. Review the KeySet produced and/or limit the acceptable characters. We will need to review the KeySet since not every time these used will be parsed here.

Either way there are two categories of chars to exclude:
1. Symbols like "-" that are problems for downstream systems like metrics. Also colon, astericks, parentheses, and brackets may be problematic.
2. Always convert to lower case. Do we want to officially not care about exact matching of case? Or, should names with UpperCase be illegal?

dave2wave:
We will analyze ConfigModels for parameter name consistency.

dave2wave:
[...]

Issue #1187, opened 2023-03-24
Configuration Layering should be implemented in a uniform way


jshook:
There is a better way to attack this. We should have a singular capability which allows us to reference external values, whether they be from the environment, other files, or whatever.
For example, a format like `{resolver:name}` can make it very clear what the resolver is. You could access an environment variable like `{ENV:HOME}`, or the contents of a file as `{FILE:myfile}`. As long as the beginning contains the resolver name, then it is uniformly parsable.

By having a singular syntactical form we can properly parse it and thus introduce the ability to layer or indirect where users may want this. It also allows us to extend the value resolver mechanism in a consistent way where each one is documented. The precedence can be established, and potentially overridden by the user.

(was) All key configuration mechanisms should work harmoniously in any layering arrangement, and there should be unit and integration tests for all of the common cases.

dave2wave:
For NBIO we have a resolver chain to sort the order for finding files in a stack of FILE, CLASSPATH, CACHE, and URL.

What does the Parameter resolution stack look like:
1. Environment variables.
2. CMD Params argvs including Atfiles.
3. Workload Params.
4. Driver Adapter and scenarios.
We resolve from inner to outer and the convention for TEMPLATEs is to take care to avoid parameter name collisions. The resolution of names and values does not go through a cohesive code path.
Without a clear plan for parameter resolution that exists in a singular place this issue cannot be implemented.

dave2wave:
The current layers look like:
1. Activity / CMD Parameters / Scenarios
2. TEMPLATE Parameters (which may or could pull in ENV variables)
3. YAML - Doc, Block, Ops

dave2wave:
There are two places that this could be applied.

1. When parsing cli parameters in ActivityDef. In this case the resolver form can make sense.
2. When parsing TEMPLATE formats in StrInterpolator. In this case the resolver format conflicts with the option to use this resolver format.

I suggest that the resolver be limited to command line parsing to expand or otherwise fill in values and in this case we should consider using the usual format for ENV and AtFiles for FILE.

dave2wave:
I'm closing this as it is contrary to current similar initiatives and it's unclear how this yet another way is an advantage

Issue #1182, opened 2023-03-22
Parameters not propagated to the op


ivansenic:
**Describe the bug**
As suggested to @jshook, I tried to use the following to define space in all operations:

```
bindings:
  space: HashRange(1,<<connections:20>>); ToString();

params:
  space: >-
    {space}

blocks:
  schema:
    ops:
      my-op:
```

However, this is not working and space needs to be defined explicitly:

```

blocks:
  schema:
    ops:
      my-op:
        space: "{space}"
```

**To Reproduce**
Setup the workflow yaml as above.

**What was Expected**
Space parameter does not have to be defined in every op explicitly.

jshook:
This may be because the params block doesn't (or at least did not) support dynamic values. As part of fixing this, the params block may need to support dynamic values. If that is not a trivial change, then and error should be thrown when a dynamic value is found in a params block.

dave2wave:
I think you've been touching related code.

dave2wave:
We'll need to throw an error if the parameter looks like a dynamic binding. Parameters are configuration while bindings are dynamic and these simply are not to be mixed.

We would need to add scanning, but it may be perfectly valid that the parameter contains a string that is used during activities. This would be up to the implementation of the driver adapter and its ops mapper.

dave2wave:
The binding format should only work if the driver adapter supports it.

PR #2104, opened 2024-12-04
Refactor ConfigModel Construction in StandardActivity

dave2wave:
This will have a refactoring of StandardActivity to provide proper scoping of ConfigModel construction along with providing advice and analysis about parameters.

We will try to solve all of these in this PR:

- [x] #1808 
- [x] #1609
- [x] #1187
- [x] #1182
- [ ] #1003
- [ ] #451
- [ ] #117
- [x] #21


dave2wave:
--advisor=validate or --parameter-model will be used to show activity, template, and model parameters use.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/engine/activityimpl/ActivityDef.java
```diff
@@ -230,4 +242,65 @@ public NBLabels auxLabels() {
...
 
     }
+
+    public NBConfigModel getConfigModel() {
+        ConfigModel cfgmodel = ConfigModel.of(this.getClass());
+        Map<String, String> params = parameterMap.getStringStringMap();
+        params.forEach((k, v) -> {
+            cfgmodel.add(Param.defaultTo(k, v, "activity parameter found on command line"));
+        });
+        if ( ! params.containsKey(FIELD_ALIAS)) {
```
jshook, line 252: I'm not entirely sure what this section should be doing with all the conditionals. This is what I'd expect: 1) individual config models are declarative and definitive about something specific, so no conditions are needed on a per-param basis. 2) config models can be derived by disjunction or intersection with others. 3) defaults in the config model are manifested when the config model is asked for the parameter that supports it.

But I might be missing something here. Can you elaborate a bit on what the purpose of this section is?

jshook (commented):
There are some nuances here that I would like to understand better.

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/engine/activityimpl/ActivityDef.java
dave2wave: `ConfigModel.add` adds keys to a model regardless of whether or not the value has already been added to the model. If the field was already by the `params.forEach` loop then it is skipped. A better fix might be to change `ConfigModel.add` behavior.

The usefulness of having a ConfigModel for ActivityDef is still speculative, but will be more for checking error conditions with excess.

dave2wave: I've adjusted the behavior in `ConfigModel.add`

jshook (dismissed):
LGTM, Thanks for digging into this!

MarkWolters (approved):
rubber stamping

PR #2107, opened 2024-12-10
Jshook/verification

jshook:
This PR contains a few key changes, and the commits are generally organized around these themes:

- fixed bug with refkey and CQL prepared statements
- API refactoring to make space functions uniform through the mapping and dispensing layers
- Fix Milvus adapter to build cleanly when enabled
- Add support for modifying op behaviors, like verification
- removed usage of the `var` keyword
- added maintainers to pom
- improved some docs

A separate PR will follow which enables the use of verification in a simpler way, for large workloads.


nb-adapters/adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/Cqld4DriverAdapter.java
```diff
@@ -29,10 +32,7 @@
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
dave2wave, line 35: This seems lazy to me when Idea will fill these in for you.

nb-adapters/adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/opmappers/Cqld4CoreOpMapper.java
```diff
@@ -61,8 +61,8 @@ public OpDispenser<Cqld4BaseOp<?>> apply(NBComponent adapterC, ParsedOp op, Long
             case raw, simple, prepared, batch -> new Cqld4CqlOpMapper(adapter).apply(adapterC, op, spaceF);
             case gremlin -> new Cqld4GremlinOpMapper(adapter, target.targetFunction).apply(adapterC, op, spaceF);
             case fluent -> new Cqld4FluentGraphOpMapper(adapter, target).apply(adapterC, op, spaceF);
-            case rainbow ->
-                new CqlD4RainbowTableMapper(adapter, spaceF, target.targetFunction).apply(adapterC, op, spaceF);
+//            case rainbow ->
+//                new CqlD4RainbowTableMapper(adapter, spaceF, target.targetFunction).apply(adapterC, op, spaceF);
```
dave2wave, line 65: I see you are commenting out rainbows. Are unicorns next? :-D

nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityconfig/yaml/OpData.java
```diff
@@ -29,47 +31,61 @@ public class OpData extends OpTemplate {
...
         this.tags = tags;
         this.bindings = bindings;
         this.params = params;
         this.op = op;
+        this.refkey = refkey;
     }
 
-    public OpData() {}
+    public OpData() {
+    }
```
dave2wave, line 50: Please explain this new `refkey`

dave2wave (approved):
LGTM (as far as one can tell with such a large PR)

nb-adapters/adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/Cqld4DriverAdapter.java
jshook: There were some places where it didn't, and I put a wildcard in to resolve issues. I thought it would just clean them up.

nb-adapters/adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/opmappers/Cqld4CoreOpMapper.java
jshook: We're going to add them back in together.

nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityconfig/yaml/OpData.java
jshook: refkey was a feature added to support efficient caching structures with the multi-client logic before, but it had a fatal error in the first edition, which was that it was being set _statefully_ on something that was lazily derived from skeleton data. (Once off the stackframe, any changes disappeared, thus prepared statements were probably broken except in the case where you had exactly 1 statement under test)

Refkey is intended to be ephemeral to an op template. It is simply a unique numeric identifier which can be used to memoize anything related to the op template in a caching structure. It is specifically an int to allow atomic reference arrays to be used. So, worst case storage is ary[] per space, and there is no hashing needed. Thus, each CQL space has its own sparse array of prepared statement referents, managed atomically and lazily. The same caching logic is used in the space caching implementation as well.

jshook: The change here was to move the refkey to the skeleton data from the yaml, added immediately by the loader as an enumeration, and exposed in the higher-order op template types as an intrinsic property

Issue #1985, opened 2024-07-19
Users should be able to specify sensitive activity parameters like passwords indirectly and uniformly.


jshook:
We have the convention of supporting userfile as well as username and passfile as well as passname, but these are spot treatments in each place they are enabled. There should be a way to do this for any activity parameter using a basic syntax.

Atfile `@somefile.yaml` support can help here, but this is not as simple as it could be. The `@` character is often annoying to deal with in shells. Instead, we may want to support a prefix on the value such as `from:` which will work for any possible value.

dave2wave:
Would a property file and then injection via `${myprops.password}`be a better approach? The property file key/values can be attached to the NBEnvironment class and be interpolated in the same places easily. In workloads I think that `TEMPLATE(myprops.password)` and `<<myprops.password>>` would work. I think that having these as global and cumulative would be an ok UX.

dave2wave:
This `user/password` pattern is only found in `Cqld4Space.java` and `Neo4JSpace.java`. `DataApiSpace.java` uses `endpointfile`. Several of other adapters use `token_file` and `token`.

It would be good to have a utility method. We should discuss @jshook

dave2wave:
One suggestion to solve this is to allow the password / token parameters to retrieve the value using a prefix like "file:<path_to_token_file>". This is complementary with #375

dave2wave:
The other part is making sure we can never expose these parameters in logs.

Parameter names interface in `ActivityDef` can somehow know which type of parameter. We need to know the purpose of the "`ToString`" and then know when to mask the value.

dave2wave:
Part of the Advisory function should be to redirect users towards using the preferred option.

dave2wave:
The ask of this issue is still nebulous:
1. Is this about adding support for properties files like "AtFiles"?
2. Is this about @ being difficult? If so, then `include=` is an alternative.
3. Is this about allowing any `value` in a key-value command line argv being expandable with the value found in a file?

dave2wave:
4. Is this about "knowing" which parameters are sensitive and making sure that these are not easily exposed in logging, etc?

Issue #1076, opened 2023-02-09
Provide a way to access values from property files.


jshook:
There should be a distinct way to access named properties from specific property files. 

This should work much like environment variable access, with similar but distinct syntax.

- Users may want to provide property files which set global System properties for later reference.
- Users may want to provide a property file specifically within a var reference, or a symbolic property file via another variable.

dave2wave:
It looks like the NBEnvironment class has the capability to take a property map in the interpolate call. It would be interesting to add a property file reference to NBAtFile or add a property file reference that is similar to AtFiles.

dave2wave:
Possible implementation would be to add a global option to NBCLIOptions which processes a property file into the NBEnvironment singleton.

dave2wave:
Do this through AtFile mechanism via suffix matching

dave2wave:
I have code to use properties files as AtFiles and just need to add tests.

As far as using properties files as environment variables the best approach is to use a global cli option to add properties through NBEnviroment.

dave2wave:
I'm going to expand the scope of this issue to support the inclusion of property files in the workload yaml.

PR #2115, opened 2024-12-14
Add Property File Support plus Related Workload Improvements

dave2wave:
closes: #1076 
closes: #1985 

1. Support the use of property files as AtFiles.
2. Support the use of property files for CmdArgLists.
3. Support the use of property files as a property overlay when interpolating strings.
4. Support the use of property files for Inclusion in workload files.
    - Also include yaml and json file inclusion.


dave2wave:
Need to adjust the insert file logic to use `NBIO` and search paths.

dave2wave:
I may add unit test, but this is ready for a review.

Issue #1805, opened 2024-02-05
Exceptions should include "context details" when provided


jshook:
It is useful to have diagnostic data provided directly from an exception that would otherwise require a user to manually dig into related configurations or other data sources.

Examples of this:
* The details for a database which returned an error.
* The UTC timecode of the error.
* Client system profile (CPUs, memory, ..)

Other details like java details, like heap utilization, etc. should generally be left out of such info, since this is properly metrics data which is provided in other places.

Because error handling mechanisms and semantics must remain robust, any mechanism which adds context to an exception must not throw any errors itself.

A new decorator type NBErrorContext should be added which indicates it can add error details in the form of a `Map<String,String>`. When an exception is thrown, the main error handling logic should walk the stack to find any such instances, and if present, the exception should be appended with ` ;; additional context: ` and the details of the error.

jshook:
Elements which have important context for interior stack frames should be decorated with this. This includes activities, or spaces which are invoked during the execution of an operation.

dave2wave:
1. Details of the tenant come from the driver. Is tenant enough for Astra or is there more metadata like region?

dave2wave:
2. Can we indicate if the error is client or serverside?
3. UTC time should always prepend the ERROR. Make it look more like an access log.

jshook:
The details found in this way need to be added to the status.yaml

jshook:
There are some open concerns on this. We should discuss them before deciding to implement anything specific here.

dave2wave:
See the discussion in #1988

dave2wave:
@jshook It's all very cool to discuss a decorator as abstract error context. It's not going to help with any implementation without a concrete example of context to add to an error that is easy to throw. Please spell this out.

dave2wave:
This is really about handling of Adapter and Op related errors in a concise manner.

PR #2128, opened 2025-01-02
Jshook/nb523

jshook:
This branch has substantial changes which aim to solve a few design issues in NB and enable much-needed testing modes. The changes are generally organized by commit, and some commits are noisier than others.

Here are the broad themes:
1. Space API formalization and alignment
2. Removal of Activity type variants, reducing to just _Activity_
3. Removal of ActivityDef, consolidating to NBConfigurable and related APIs
4. Complete overhaul of the Op Synthesis layer, which streamlines and clarifies op synthesis logic.
5. Enable nascent field capture API
6. Initial work on a very simple-to-configure database verification layer. This will be elaborated once the changes above are stabilized.
7. Introduction of basic composable op behaviors, as part of field verification, etc.
8. Retooling onto Java 23

Due to the scope of these changes, we should review and shore-up test coverage where possible. I'm pushing this up for visibility and to get more hands on it sooner.


nb-adapters/adapter-cqld4/src/main/java/io/nosqlbench/adapter/cqld4/optypes/Cqld4CqlOp.java
```diff
@@ -169,7 +169,7 @@ public CycleOp<?> getNextOp() {
     }
 
     @Override
-    public Function<List<Row>, Map<String, ?>> initCaptureF(CapturePoints<List<Row>> points) {
+    public Function<List<Row>, Map<String, ?>> captureF(CapturePoints<List<Row>> points) {
```
jshook, line 172: This will be moved to the dispenser layer

nb-adapters/adapter-diag/src/main/java/io/nosqlbench/adapter/diag/DiagOpDispenser.java
```diff
@@ -102,6 +103,16 @@ public NBConfigModel getReconfigModel() {
         return opFunc.getReconfigModel();
     }
 
+    @Override
+    public void applyConfig(NBConfiguration cfg) {
+
```
jshook, line 108: missing impl details, double check test coverage

nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityconfig/yaml/OpDef.java
```diff
@@ -105,7 +105,6 @@ public Map<String, String> getTags() {
     private LinkedHashMap<String, String> composeTags() {
         LinkedHashMap<String, String> tagsWithName = new LinkedHashMap<>(new MultiMapLookup<>(rawOpDef.getTags(), block.getTags()));
         tagsWithName.put("block",block.getName());
-        tagsWithName.put("name",this.rawOpDef.getName());
```
jshook, line 108: This removal should be explained better

nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityconfig/yaml/OpTemplates.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.*;
+import java.util.stream.Stream;
+
+/// [OpTemplates] is a list of selected op templates and their backing data.
+///
+/// It is a value type which makes it easy to /// get matching subsets of op templates according to tag filters, to combine them, etc.
```
jshook, line 33: A formatting pass should be made over all these files after the merge

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/components/core/NBBaseComponent.java
```diff
@@ -96,24 +102,47 @@ public NBComponent getParent() {
...
-            this.children.add(child);
+            this.children.add(adding);
         }
         return this;
     }
 
     @Override
     public NBComponent detachChild(NBComponent... children) {
+        for (NBComponent child : children) {
+            logger.debug(() -> "notifyinb before detaching " +
```
jshook, line 140: user-visible typo

nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/components/core/NBComponentExecutionScope.java
```diff
@@ -29,7 +29,9 @@ public NBComponentExecutionScope(NBComponent... components) {
     @Override
     public void close() throws RuntimeException {
         for (NBComponent component : components) {
-            component.beforeDetach();
+//          This is now handled inline with [NBComponent#detachChild], which puts it after the
+//          out of scope notification -- this might need testing adjustments or clarification
```
jshook, line 33: audit testing coverage for this

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityapi/core/IActivityWiring.java
```diff
@@ -0,0 +1,64 @@
+/*
```
jshook, line 1: change IInterface naming to be consistent with others

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityapi/planning/DerivedSequence.java
```diff
@@ -0,0 +1,81 @@
...
+/// based on a mapping function. This is done lazily by default, to allow for incremental
+/// initialization.
+public class DerivedSequence<BASE, DERIVED> implements OpSequence<DERIVED> {
+    private final OpSequence<BASE> baseTypeSequence;
+    private final SequencerType type;
+    private final List<DERIVED> elems;
+    private final int[] seq;
+    private final Function<BASE, DERIVED> deriveF;
+
+    public DerivedSequence(OpSequence<BASE> baseTypeSequence, Function<BASE,DERIVED> deriveF) {
```
jshook, line 35: This class needs better explanation and probably testing coverage too

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/AdapterResolver.java
```diff
@@ -0,0 +1,62 @@
...
+    )
+    {
+        ServiceSelector<DriverAdapterLoader>
+            loader =
+            ServiceSelector.of(name, ServiceLoader.load(DriverAdapterLoader.class));
+        DriverAdapterLoader
+            dal =
+            loader.get()
+                .orElseThrow(() -> new OpConfigError("No DriverAdapterLoader found for " + name));
+        DriverAdapter<CycleOp<?>, Space> adapter = dal.load(parent, NBLabels.forKV());
```
dave2wave, line 52: Several comments here:
1. I don't like this code style with unneeded multiple lines.
2. Call the `loader` variable `selector`.
3. Call the `dal` variable `loader`.
4. Use a version of the "helpful" config error.
   ```java
    //        if (defaultDriverName.isPresent() && defaultAdapter.isEmpty()) {
    //            throw new BasicError("Unable to load '" + defaultDriverName.get() + "' driver adapter.\n" + "Rebuild NB5 to include this " + "driver adapter. Change" + " '<activeByDefault>false</activeByDefault>' for the driver in" + " './nb-adapters/pom.xml' and" + " './nb-adapters/nb-adapters-included/pom.xml' first.");
    //        }
   ```

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/OpResolution.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /// Converting an op template into a ParsedOp means back-filling it with config data from the
+    /// activity parameter.
+    public synchronized <OPTYPE extends CycleOp<?>, SPACETYPE extends Space> ParsedOp getParsedOp() {
+        if (resolvingParsedOp == true) {
+            throw new OpConfigError("Auto-recursion while resolving dispenser for op '" +
+                                    template.getName() +
+                                    "'");
+        }
```
dave2wave, line 143: See previous comment

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/OpResolution.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /// Each op template is interpreted by a specific [OpMapper] designated by the `driver` op
+    /// field. Thus
+    /// the associated driver needs to be loaded.
+    private synchronized <OPTYPE extends CycleOp<?>, SPACETYPE extends Space> DriverAdapter<OPTYPE, SPACETYPE> resolveAdapter() {
+        if (resolvingAdapter) {
+            throw new OpConfigError("Auto-recursion while resolving adapter for op '" +
+                                    template.getName() +
+                                    "'");
+        }
```
dave2wave, line 160: See previous/

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/Activity.java
```diff
@@ -0,0 +1,969 @@
...
+        //        this.metrics = new ActivityMetrics(this);
+
+        //        OpsDocList workload;
+        Optional<String> yaml_loc = config.getOptional("yaml", "workload");
+
+        // TODO: avoid having to load this duplicitously to parse the template variables in a separate phase
+        NBConfigModel yamlmodel = yaml_loc.map(path -> {
+            return OpsLoader.loadPath(path, new LinkedHashMap<>(config.getMap()), "activities")
+                .getConfigModel();
+        }).orElse(ConfigModel.of(Activity.class).asReadOnly());
```
dave2wave, line 156: `yamlmodel` is unused. If you want to have the non-strict expansion of the TEMPLATE parameters into the Activity model then you will need to "reconfigure". This would be considered the difference between the now strict implementation and our current loose model.

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/Activity.java
```diff
@@ -0,0 +1,969 @@
...
+        //        Optional<DriverAdapter<?, ?>> defaultAdapter = activityDef.getParams()
+        //            .getOptionalString("driver")
+        //            .flatMap(name -> ServiceSelector.of(name, ServiceLoader.load(DriverAdapterLoader.class)).get())
+        //            .map(l -> l.load(this, NBLabels.forKV()));
+        //
+        //        if (defaultDriverName.isPresent() && defaultAdapter.isEmpty()) {
+        //            throw new BasicError("Unable to load '" + defaultDriverName.get() + "' driver adapter.\n" + "Rebuild NB5 to include this " + "driver adapter. Change" + " '<activeByDefault>false</activeByDefault>' for the driver in" + " './nb-adapters/pom.xml' and" + " './nb-adapters/nb-adapters-included/pom.xml' first.");
+        //        }
+
+        //        NBConfigModel supersetConfig = ConfigModel.of(Activity.class).add(yamlmodel);
```
dave2wave, line 225: This is the former use of `yamlmodel`,

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/Activity.java
```diff
@@ -0,0 +1,969 @@
...
+        //        List<ParsedOp> activeParsedOps = ts.filter(allParsedOps);
+
+        //        if (defaultDriverOption.isPresent()) {
+        //            long matchingDefault = mappers.keySet().stream().filter(n -> n.equals(defaultDriverOption.get())).count();
+        //            if (0 == matchingDefault) {
+        //                logger.warn(
+        //                    "All op templates used a different driver than the default '{}'",
+        //                    defaultDriverOption.get()
+        //                );
+        //            }
```
dave2wave, line 245: In the new object model it's unclear to me how to make this test. Perhaps by looking into expanding how `adapterCache` map functions.

nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityimpl/FieldVerifier.java
```diff
@@ -0,0 +1,235 @@
...
+        this.resultsVerifiedError = parent.create().counter(
+            "results_verified_error", MetricCategory.Verification,
+            "The number of results which have been verified with no error"
+        );
+        this.resultsOkCounter = parent.create().counter(
+            "results_verified_ok",
+            MetricCategory.Verification,
+            "The number of results which had " + "a verification error"
+        );
+        this.verifiedFieldsCounter = parent.create().counter(
```
MarkWolters, line 56: What is the utility in tracking this?

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/AdapterResolver.java
jshook: Fully agreed on 1-3.
For 1) I was experimenting with auto-formatting to address some of the inconsistencies and this was a format from an earlier setting which needs to be updated. I think we'll need settle on a formatting standard separately from this PR, and I'm eager for us to do that to ease PRs and reviews.

On 4) I'm ok with that prescriptive error as long as we don't send users down a false path. (i.e, they just had a typo, and no such adapter exists)

nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityconfig/yaml/OpDef.java
dave2wave: name is just a synonym for ops

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/AdapterResolver.java
dave2wave: Sure - we can approach 4) in other ways including documentation.

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/OpResolution.java
```diff
@@ -0,0 +1,174 @@
...
+    /// to get the components or component functions around which to build the final dispenser.
+    /// These elements may be
+    /// initialized by side-effect of other operations being resolved. In every case, once an
+    /// element of any op
+    /// resolution is resolved, it should be considered _defined_, and not resolved again.
+    public synchronized <OPTYPE extends CycleOp<?>, SPACETYPE extends Space> OpDispenser<OPTYPE> resolveDispenser() {
+        if (resolvingDispenser == true) {
+            throw new OpConfigError("Auto-recursion while resolving dispenser for op '" +
+                                    template.getName() +
+                                    "'");
```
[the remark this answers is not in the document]
jshook: Yes, this is expected to manifest only when a user specifies an op template which causes infinite recursion.

Example Scenario:
A user wants to configure a database verifier op by saying 'take all fields and expected values from other op indicated by tag filter=...', and the tag filter is wrong. For example, if you specify a tag filter which matches the current op (a self-referencing op template), then it would cause a dependency loop.

Perhaps _auto-recursion_ is probably not the best description of this problem, accurate or not. Honestly, for possible issues like this which are more than a simple surface-level mistake, I think we should have clearly defined error monikers which point to troubleshooting docs. (In that case op-template-cyclic-reference might be a reasonable one for this)

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/Activity.java
jshook: Yes, the adapter cache is a product of op resolution, and can be checked for canonical mapping of active adapters. We can make sure this is exposed. We might want to do this where we isolate a particular verification checkpoint where we assert sanity on the assembled op dispensers.

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/OpResolution.java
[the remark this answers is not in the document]
dave2wave: So, the situation is a recursive filter? Is it possible to automatically modify or add to the filter such that this recursion is blocked? Or, should the appropriate filter change be suggested by the error information. That would be in the spirit of #1805

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/Activity.java
dave2wave: adapterCache includes the default Adaptor. Instead scan the opTemplates into an adapter list and check vs. the default adapter.

MarkWolters (approved):
I think this all looks very good. Could use a tour of some of the core reworking around Action at some point. Thank you for including the diagrams, I think those really add some value.

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/OpResolution.java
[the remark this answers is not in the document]
jshook: Yes, a recursive filter or possibly another lookup mechanism besides a filter match. In the case that the user specfies a filter which causes a cyclic reference, it is not clear what the reason for the error is, and I'm not sure we can make any safe assumptions here. We can possibly suggest reasons and update the error output as we see it occur.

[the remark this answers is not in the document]
jshook: Let me say that better: It is not clear what the user was intending to effect, thus it is not clear what guidance to provide.

nb-engine/nb-engine-core/src/main/java/io/nosqlbench/engine/api/activityimpl/uniform/Activity.java
jshook: adapterCache should not include the default unless it was referenced by an op template. Also, I'm beginning to think this is not really an issue, since there are plenty of cases where a default adapter might be supplied and not used (test automation)

jshook: I'd like to take this scope as a follow-on for this PR if that's ok.

nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityimpl/FieldVerifier.java
```diff
@@ -0,0 +1,235 @@
...
+                List<String> vars = referenced.get().getBindPoints().stream().map(
+                    bp -> bp.getAnchor()).toList();
+                fields.addAll(vars);
+                bindings.addAll(vars);
+            } else {
+                throw new OpConfigError(
+                    "no op found for verify setting '" + fieldSpec + "' " + "for op " + "template" + " '" + pop.getName() + "'");
+            }
+        } else {
+            String[] vfields = fieldSpec.split("\\s*,\\s*");
```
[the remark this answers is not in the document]
jshook: I'll add docs here

nb-apis/adapters-api/src/main/java/io/nosqlbench/adapters/api/activityimpl/FieldVerifier.java
```diff
@@ -0,0 +1,235 @@
...
+                if (missingRowFields.size() > 0) {
+                    diff += missingRowFields.size();
+
+                    logbuffer.append("\nexpected fields '");
+                    logbuffer.append(String.join("','", missingRowFields));
+                    logbuffer.append("' not in row.");
+                }
+            }
+
+//            if (diffType.is(DiffType.rowfields)) {
```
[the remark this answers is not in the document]
jshook: temporary

Issue #2146, opened 2025-01-15
verification timer should be separate from execute timer


jshook:
In StandardAction, the execute timer includes the verification timer fully. These should be serialized in terms of wall clock time.

PR #2147, opened 2025-01-15
nosqlbench-2146 verification timer should be separate from execute timer

jshook:
moves verify timer out of execute timer
closes: #2146


michaeljmarshall (approved):
LGTM, thanks for the quick patch!

PR #2159, opened 2025-02-05
Mwolters/pred enhancements

MarkWolters:
new binding functions to work with simple predicate datasets in hdf5


nb-virtdata/virtdata-lib-hdf5/src/main/java/io/nosqlbench/virtdata/library/hdf5/from_long/to_string/IntArrayToString.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.function.Function;
+
+@ThreadSafeMapper
+@Categories(Category.experimental)
+public class IntArrayToString implements Function<int[],String> {
+    private final StringBuilder sb = new StringBuilder();
+
+    @Override
+    public String apply(int[] ints) {
+        sb.setLength(0);
```
jshook, line 32: This is not thread-safe. You can just instance sb within the method to put it on the stack. I've often used ThreadLocal to provide thread-specific buffers like this in the past, but lately it's not as much of an improvement over just making a new buffer when needed.

jshook (changes requested):
Only one issue needs attention before merge

nb-virtdata/virtdata-lib-hdf5/src/main/java/io/nosqlbench/virtdata/library/hdf5/from_long/to_string/IntArrayToString.java
MarkWolters: updated in latest checkin

jshook (approved):

PR #2161, opened 2025-02-18
distribution morphing support

jshook:
This will allow us to accurately test degree distributions and even distributions or something in between if needed


MarkWolters (approved):
Looks good. Need to think a bit about relating this to the query side but I don't see a problem with merging this to use for data population, interesting technique!

PR #2162, opened 2025-02-19
Update CQL Java Driver to 4.19.0

msmygit:
Update CQL Java Driver to [`4.19.0`](https://github.com/apache/cassandra-java-driver/releases)


jshook (approved):
the java driver deps should be in the java driver module. Not sure why it is there in the reactor project root pom. Let's let it ride for now to get it updated, and then we can move it after making sure it's not going to break something.

PR #2164, opened 2025-02-27
Binpred

MarkWolters:


nb-virtdata/virtdata-lib-hdf5/src/main/java/io/nosqlbench/virtdata/library/hdf5/from_long/to_string/predicate_parser/BinStringToCql.java
```diff
@@ -0,0 +1,35 @@
...
+package io.nosqlbench.virtdata.library.hdf5.from_long.to_string.predicate_parser;
+
+import io.jhdf.api.Dataset;
+
+public class BinStringToCql extends BinToCqlProcessor {
+    private final String[] predicateArray;
+
+    public BinStringToCql(Dataset predicateDataset, String[] schema) {
+        super(schema);
+        predicateArray = (String[]) predicateDataset.getData();
```
jshook, line 27: Not sure if this is needed, as in not sure if String format should be valid here for bin encoding.

jshook: Let me be more specific. I think that a string decoder would need to be based on a string tokenization, and thus would not fit in the binary view of decoding.

nb-virtdata/virtdata-lib-hdf5/src/main/java/io/nosqlbench/virtdata/predicates/nodewalk/types/Node.java
```diff
@@ -0,0 +1,5 @@
+package io.nosqlbench.virtdata.predicates.nodewalk.types;
+
+/// [T] is the node type, from
+public sealed interface Node<T> extends BBWriter<T> permits ConjugateNode, PredicateNode {
+}
```
jshook, line 5: There was a bug in the nbvectors version of this which was fixed, as in you couldn't reliably use ConjugateNode or PredicateNode to load an arbitrary predicate, since the decoding is specialized for each. So a static method was added to the base Node class which would pre-read the node type and delegate thusly. This probably needs to be pulled over.

nb-virtdata/virtdata-lib-hdf5/src/test/java/io/nosqlbench/virtdata/predicates/nodewalk/repr/H5JsonNodeRendererTest.java
```diff
@@ -0,0 +1,128 @@
+package io.nosqlbench.virtdata.predicates.nodewalk.repr;
```
jshook, line 1: Nice tests!

jshook (changes requested):
I think this will be ready for merge once we address the updates/bug fixes in the original encoder/decoder

nb-virtdata/virtdata-lib-hdf5/src/main/java/io/nosqlbench/virtdata/library/hdf5/from_long/to_string/predicate_parser/BinStringToCql.java
MarkWolters: It's not I suppose, I'm just being a hoarder.

MarkWolters: I can take it out

nb-virtdata/virtdata-lib-hdf5/src/main/java/io/nosqlbench/virtdata/predicates/nodewalk/types/Node.java
MarkWolters: Done

nb-virtdata/virtdata-lib-hdf5/src/main/java/io/nosqlbench/virtdata/library/hdf5/from_long/to_string/predicate_parser/BinStringToCql.java
MarkWolters: I see your point. IMO it is valid to have a string processor here, but the way the data is being stored in the hdf5 file in that case should be an actual string encoding, and not binary data that hdf5 is treating internally as a string, which is essentially "wrong" on the encoding end. So I'm going to leave the string processor here but only as "not yet implemented" functionality. Let me know if that works for you.

jshook (approved):
LGTM, This will be cool

PR #2166, opened 2025-03-06
Mwolters/metadata

MarkWolters:
1. Added dependency for nbvectors library published from nbdatatools and removed duplicate classes defined in nb 
2. Added attribute processing for determining data type from metadata if available
3. general cleanup and comments


jshook (approved):
Nice

PR #2172, opened 2025-04-04
fix for large fvec/ivec files

MarkWolters:


nb-virtdata/virtdata-lib-hdf5/src/main/java/io/nosqlbench/virtdata/library/ivecfvec/FVecReader.java
```diff
@@ -37,58 +35,103 @@
...
         int recordIdx = (int) (value % reclim);
-        int recpos = recordIdx*reclen;
-        int recdim = Integer.reverseBytes(bb.getInt(recpos));
-        if(recdim!=dimensions) {
-            throw new RuntimeException("dimensions are not uniform for fvec file '" + this.path.toString() + "', found dim " + recdim + " at record " + value);
+        long recpos = (long)recordIdx * reclen;
+
+        try {
+            // Read record dimensions
+            headerBuffer.clear();
```
jshook, line 106: This isn't thread-safe, and neither is the other buffer. The same single-instanced function will be shared by many threads, so you might want to add a test to show that.
Allocating bytebuffers, and arrays in particular is quite fast now, so it don't hurt to just make one for each call. GC can handle it much better now too.

jshook (changes requested):
Unless I'm mistaken (and I could be!) I think all of the buffers need to be on the stack except for reference buffers (like the one to ensure dim is constant).
The file channels can't use relative positioning consistently between threads, but they might be ok with absolute positioning.

nb-virtdata/virtdata-lib-hdf5/src/main/java/io/nosqlbench/virtdata/library/ivecfvec/FVecReader.java
MarkWolters: I think the concurrency issues you point out have been fixed and tested in the latest checkin. Please review at your convenience and if you still have concerns let me know and I'll take care of it.

jshook (approved):
I think this looks ok. I'm not super familiar with the semantics of scope access, but the testing seems good enough to pull this in.

PR #2175, opened 2025-04-09
Jshook/recall 2173

### LINKED ISSUE #2173: recall compute functions are not correct for k<n(gt)
When the result size or top-K is less than the size of the ground truth, the recall (and possibly other) functions do not use the correct divisor or truncation value.

Fixes should include:
* set truncation should be controlled by k explicitly
* set truncation should only happen only after ordering
* the divisor should simply be k, and not implicit based on gt or result size
* assertion for n(gt)>=k should be made

Previous testing which has been based on using the whole ground truth should not be affected, given that:
* gt was already ordered
* when k==n(gt), the correct value is computed

jshook:
Includes fix for recall from #2173
commits are itemized around smaller changes to simplify review if needed
closes #2173


tlwillke:
The PR looks good overall.  Are all of the included tests passing?

tlwillke (approved):
+1

PR #2187, opened 2025-06-06
PR: nb5-jdbc-bug-fix

MikeYaacoubStax:
This only fixes the bug with the prepared Statement ReadOp.
I added a simple workload and a more complicated one that fails because of the prepared statement caching bug. More to come.


jshook (approved):
lgtm

PR #2188, opened 2025-06-08
PR: Opensearch dusting off

MikeYaacoubStax:
NB already supported OpenSearch via an AWS Opensearch adapter, I extended it to support a non-AWS open source version deployment of Opensearch (to use this, you can simply provide the driver params `host` and `port`). For example:

`nb5 opensearch_basic.yaml driver=opensearch host=localhost port=9200`

The additional workloads have been tested with a local deployment of an Opensearch DB.


MikeYaacoubStax:
Moved this to another branch.

PR #2193, opened 2025-06-18
Bvecsupport

jshook:
This makes bvec (and xvec) support more modular and fixes an issue with testing scope.


nb-virtdata/virtdata-lib-io/src/main/java/io/nosqlbench/virtdata/library/xvec/BVecToFloatReader.java
```diff
@@ -51,7 +51,7 @@ public BVecToFloatReader(String filePath) throws IOException {
             throw new RuntimeException(e);
         }
         dimBuffer.flip();
-        this.dimension = Integer.reverseBytes(dimBuffer.getInt());
```
MarkWolters, line 54: so dimension is stored in little endian and you are updating to not reverse it? Does that not result in vector size being calculated wrong?

PR #2238, opened 2025-12-04
Data api client bump to 2.x + ensure find-operations exhaust the cursor before completing op

### LINKED ISSUE #2237: Data API adapter: upgrade to client 2.x + mishandled 'find' operations
**Describe the bug**

In the Data API adapter, two of the three "find" ops (FindVectorFilter and FindVector) do not actually end up doing the API request. The reason is that in the Data API client, just calling `collection.find` only prepares a lazy iterator, which then must be exhausted for the requests to be made.

Additionally: the Data API client released 2.x, which comes with (a few breaking changes and) extended scope and coverage for more commands. For instance, it can cover HCD better, supports findAndRerank, binary-encoded vectors and so on.

It might be convenient to do both (the bugfix and the version bump) at once, because the find/iterator API in clients is one of the parts that changed significantly between 1.x and 2.x.

**To Reproduce**

Measuring the "result

" metric on an op like

```
  thick_find_ann:
    ops:
      my_find_op:
        find_vector_filter: "TEMPLATE(collection)"
        vector: "{seq_insert1_vector}"
```

reports something order of ~ 0.1 ms (percentiles & max) when run against Astra. This is just the time needed to instantiate the cursor by the client.

**What was Expected**
One would expect something O(100 ms) or so from the above op.

**Additional context**

Verified with `5.25.0-release`.

**Screenshots, if applicable**

The attachment is a plot of the metric exemplified above. The rightmost point, which gets the right (aka expected) order of magnitude, is from a build which I had fixed to actually exhaust the cursor as part of the op's `apply` before it returns.

Image

sl-at-ibm:
This PR ports the data-api adapter to using the latest client (2.1.x) and fixes an incorrect behaviour of some of its `find` operations which otherwise return before any API communication is ever initiated, making the response time measurements invalid.

Fixes #2237 .

### DIFF
--- a/nb-adapters/adapter-dataapi/pom.xml
+++ b/nb-adapters/adapter-dataapi/pom.xml
@@ -49,7 +49,7 @@
     <dependency>
         <groupId>com.datastax.astra</groupId>
         <artifactId>astra-db-java</artifactId>
-        <version>2.1.2</version>
+        <version>2.1.4</version>
     </dependency>
     <!-- Override Lombok version for Java 25 compatibility -->
     <dependency>
--- a/nb-adapters/adapter-dataapi/src/main/java/io/nosqlbench/adapter/dataapi/opdispensers/DataApiOpDispenser.java
+++ b/nb-adapters/adapter-dataapi/src/main/java/io/nosqlbench/adapter/dataapi/opdispensers/DataApiOpDispenser.java
@@ -231,12 +231,12 @@ public abstract class DataApiOpDispenser extends BaseOpDispenser<DataApiBaseOp, 
         if (providerFunc.isPresent() && modeFunc.isPresent()) {
             LongFunction<String> pf = providerFunc.get();
             LongFunction<String> mf = modeFunc.get();
-            if (apiKeyFunc.isPresent()){
+            if (apiKeyFunc.isPresent()) {
                 LongFunction<String> ak = apiKeyFunc.get();
                 optionsBldr = paramFunc.isPresent() ?
                     optionsBldr.vectorize(pf.apply(l), mf.apply(l), ak.apply(l), paramFunc.get().apply(l)) :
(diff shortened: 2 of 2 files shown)


jshook (approved):
These look like API alignment and house keeping changes to me. In general the look good. I only commented on a couple things you might want to improve, but since these were not stepwise regressions in the code, I don't think the PR should be blocked.

sl-at-ibm:
I have not rebased yet, still off 5.25.0-release pending verification/resolution of issues with writing CSV found on latest main.

sl-at-ibm:
> LGTM 👍🏼 but could you also please update the files and examples under https://github.com/sl-at-ibm/nosqlbench/tree/SL-data-api-adapter/nb-adapters/adapter-dataapi/src/main/resources/activities directory as well to match the new pattern? Thank you in advance!

@msmygit Thanks for taking a look!
This PR introduces no changes to the public API (/syntax) for Data API activities, it's all internal changes (for the client ->2.x upgrade) and the `find` bugfix.
I have tested all the syntax exemplified in the directory you quoted, and the few combinations that gave errors... error on main as well right now.
So, while I agree that fixes and updates to the examples will definitely need to be the topic of upcoming work (I have detailed notes as to how to proceed and we can discuss them if you like :heart: ), I would rather not lump them into this PR which is already massive as it is now.

nb-adapters/adapter-dataapi/pom.xml
```diff
@@ -49,7 +49,7 @@
     <dependency>
         <groupId>com.datastax.astra</groupId>
         <artifactId>astra-db-java</artifactId>
-        <version>1.1.1</version>
+        <version>2.1.2</version>
```
msmygit: Should we move to `2.1.4`? https://github.com/datastax/astra-db-java/compare/2.1.2..2.1.4

```suggestion
        <version>2.1.4</version>
```

jshook (approved):
Thanks for the update!

PR #2290, opened 2026-08-06
Fast Zipf sampler implementation based on Pareto distribution

### LINKED ISSUE #91: Optimise Zipf distribution initialisation and / or warn users about its duration
The initialisation phase for the Zipf distribution with a keycount higher than 10k is very long. This is likely due to the time it takes to precompute the lookup table.

It would be great to optimise this phase so it is quicker if possible. At the very least, we should have a warning message to let users know that this phase is running and it is likely to take time to complete.

stonewhitener:
resolves #91

References

* https://dl.acm.org/doi/10.1145/191843.191886
* https://en.wikipedia.org/wiki/Pareto_distribution


stonewhitener:
Now, ready for review.

jshook (approved):
The only measurable trade-off here (minor accuracy divergence)  is a minor concern compared to how slow zipf modeling can be with certain parameters. This is a very good change. Thanks for the contribution!

PR #2291, opened 2026-08-13
Fix timezone-sensitive NBEnvironment test

stonewhitener:
The test `NBEnvironmentTest.testInterpolationWithTimestamp` failed when I ran it in Japan because the expected date was hard-coded for UTC, so I updated it to account for the JVM’s default time zone.


jshook (approved):
Thanks for the determinism

PR #2297, opened 2026-08-19
Flush pending metric snapshots at activity completion

stonewhitener:
## Summary

Fix missing final-interval metrics when an activity completes before a slower reporter's cadence bucket is full.

## Problem

`triggerSnapshot()` previously bypassed the cadence hierarchy and sent only the newly captured view directly to every consumer. This ensured that the latest delta reached reporters such as SQLite, but snapshots already accumulated in slower cadence buckets remained pending.

For delta-based histograms, those earlier observations had already been consumed during periodic capture and could not be reconstructed later. If the activity or session ended before the bucket boundary, the pending observations were never written to the database.

## Changes

- Route the final snapshot through the normal cadence hierarchy.
- Recursively flush all remaining partial buckets after the final capture.
- Synchronize periodic and explicit captures to prevent overlapping delta snapshots.
- Flush previously pending snapshots even if the final capture throws an exception.
- Add test coverage for:
  - A nested `base → 2× → 4×` cadence hierarchy.
  - Pending snapshot delivery when the final capture fails.

## Testing

```shell
mvn -Paccuracy -pl nb-apis/nb-api -am \
  -Dtest=MetricsSnapshotSchedulerTest \
  -Dsurefire.failIfNoSpecifiedTests=false test
```

Result: 8 tests passed with no failures or errors.

### DIFF
--- a/nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/engine/metrics/MetricsSnapshotScheduler.java
+++ b/nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/engine/metrics/MetricsSnapshotScheduler.java
@@ -261,6 +261,12 @@ public final class MetricsSnapshotScheduler extends UnstartedPeriodicTaskCompone
 
     @Override
     protected void task() {
+        synchronized (scheduleLock) {
+            captureSnapshotLocked();
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


jshook (approved):
I think the testing is sufficient here. This is a much needed improvement. Thanks!

PR #2298, opened 2026-08-21
Fix metrics scheduler lifecycle across rebases

stonewhitener:
## Summary

This PR fixes two related lifecycle issues when `MetricsSnapshotScheduler` rebases to a faster sampling interval:

1. The replacement scheduler could disappear from the scheduler registry.
2. Consumers registered before the rebase could not unregister from the replacement scheduler.

## Problem

Consider two reporters registered in this order:

1. Reporter A requests a 200 ms interval, creating `Scheduler A`.
2. Reporter B later requests a 100 ms interval.
3. The scheduler rebases from 200 ms to 100 ms by replacing `Scheduler A` with `Scheduler B`.

Previously, the rebase called `teardown()` from inside `schedulers.compute(...)`. That teardown modified the **same** registry, so the newly created scheduler could be returned and started without remaining registered.

https://github.com/nosqlbench/nosqlbench/blob/4a5e567718416c7949f74695ccf9d9c0db6ef1ea/nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/engine/metrics/MetricsSnapshotScheduler.java#L112-L124

https://github.com/nosqlbench/nosqlbench/blob/4a5e567718416c7949f74695ccf9d9c0db6ef1ea/nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/engine/metrics/MetricsSnapshotScheduler.java#L326-L332

There was also a stale-reference problem. Reporter A retained its original scheduler reference while its registration was copied to the replacement:

```text
Reporter A ──► Scheduler A (retired, no registrations)

Registry ────► Scheduler B (active)
               └─ Reporter A registration
```

When Reporter A closed, it called `unregisterConsumer()` on retired Scheduler A. The call became a no-op, leaving Reporter A registered with Scheduler B. The active scheduler could then continue invoking a closed reporter, potentially accessing closed CSV streams, SQLite connections, or other released resources.

## Solution

The rebase lifecycle now:

- Retires the old scheduler without modifying the registry entry currently being computed.
- Removes scheduler entries only during normal teardown, using a conditional key/value removal so an old scheduler cannot remove its replacement.
- Transfers registrations while holding `scheduleLock`.
- Publishes a successor reference atomically with the registration transfer.
- Delegates unregister calls from retired schedulers to their successors.
- Releases the old lock before delegation to avoid nested scheduler locks.

## Testing

The focused regression coverage verifies:

- `testRebasesWithSmallerInterval`
  - Ensure that `lookup(root)` returns the replacement scheduler.
- `testUnregistersThroughRetiredScheduler`
  - Ensure that an unregister call through the retired scheduler removes the consumer from the active replacement.

Validation completed successfully:

- `MetricsSnapshotSchedulerTest`: 9 tests passed.
- `MetricsReporterIntegrationTest`: 5 tests passed.

### DIFF
--- a/nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/engine/metrics/MetricsSnapshotScheduler.java
+++ b/nb-apis/nb-api/src/main/java/io/nosqlbench/nb/api/engine/metrics/MetricsSnapshotScheduler.java
@@ -120,10 +120,9 @@ public final class MetricsSnapshotScheduler extends UnstartedPeriodicTaskCompone
                     throw new IllegalArgumentException(
                         "Requested interval " + intervalMillis + " must divide existing base interval " + existing.baseIntervalMillis);
                 }
-                List<ScheduleRegistration> registrations = existing.snapshotSchedules();
-                existing.teardown();
+                existing.retireForRebase();
                 MetricsSnapshotScheduler replacement = new MetricsSnapshotScheduler(root, intervalMillis);
-                replacement.restoreSchedules(registrations);
+                existing.transferSchedulesTo(replacement);
                 created[0] = replacement;
                 return replacement;
             }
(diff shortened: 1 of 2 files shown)


jshook (approved):
NIce catch. I think it can be merged as it is, but it would also be nice to see some higher-level numerical tests which illustrate the practical effects of the bug and fix with metric data.

